Issue #6, opened 2018-02-16
Separate CLI from from library


adelavega:
Currently, the CLI and the fitlins library are very interlaced.
That is, it's currently not intuitive to use fitlins as a library that can be imported by another python program.

For example, default values for parameters come from argparse, and the function `create_workflow` takes a since opts argument, rather than explicitly stating the different inputs.

IMO, it would be useful to divorce these two a bit, so that `create_workflow` could be called without having to go through the CLI. This would make it easy for `neuroscout-cli` to wrap fitlinns without having to resort to using subprocess.

Thoughts?

adelavega:
I suppose the counterargument here is that `run.py` *is* the CLI so if I want to use fitlins as a library, I should create an equivalent function that imports from `fitlins.base` etc, but it seems that `create_workflow` does a lot of important stuff and I think it'd be nice to use `fitlins` default wherever possible.

effigies:
Sure, that's not too big a deal. The `create_workflow` is a hold-over from fmriprep, when I thought this would move quickly to a nipype workflow oriented approach.

The really fundamental bits are just the `init`, `first_level` and `second_level` in `fitlins.base`. It wouldn't be too hard to wrap those up with the 5 or 6 lines from `run.py`, but what would your ideal API look like?

adelavega:
Hard to say right now, but probably something pretty simple like:

    analysis = Analysis(**opts)

`__init__` of `Analysis` would do a lot of what `create_workflow` does right now, including loading configs, prepping in/out dirs, creating the worklow

then:
    `analysis.run(n_procs=10)` 

which runs the workflow

One downside of moving the defaults out of argsparse I suppose is that its harder for a `CLI` user to see them.

adelavega:
For now, let me poke around with reorganizing `create_workflow` a bit so I can call it from outside and we can worry more about a bigger refactor when my use case becomes more clear.

effigies:
I think the easiest thing will be to wrap these lines in a function that provides all the necessary inputs (model, bids_dir, preproc_dir, deriv_dir, level):

https://github.com/effigies/fitlins/blob/e77409e062fd841d76615612a5b43cfff58eaf93/fitlins/cli/run.py#L210-L217

You can throw that into `fitlins.base` if you want.

I'm not a huge fan of the big fat config object approach to organizing, though, but I am thinking about this from a front-end perspective, not a library perspective. Be happy to consider any proposals you've got.

adelavega:
[...]

PR #7, opened 2018-02-16
WIP: Allow importing and calling main

adelavega:
Here's on option to address #6 

First, I removed the defaults from argparse and put them in `create_workflow`. This actually ends up not mattering much but IMO its cleaner and more explicit

More importantly, `main` now takes an `args` argument which is passed to `parse_args`. 
So now you can do the followng:

    from fitlins.cli.run import main
    main('dataset', 'output_dir', 'dataset', '--model=model.json')

and argparse will parse it. If no argument is passed, command line arguments (from sys) are passed in instead.

There's probably a better way to have a library CLI but for now this is good enough for me.

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -101,10 +101,10 @@ def get_parser():
     return parser
 
 
-def main():
+def main(args=None):
     """Entry point"""
     warnings.showwarning = _warn_redirect
-    opts = get_parser().parse_args()
+    opts = get_parser().parse_args(args)
     if opts.debug:
         logger.setLevel(logging.DEBUG)
 


adelavega:
Oh also, is this a Python 3 module or Python 2/3? I guess given that its a BIDS app you don't have to care too much about Python 2 compatibility.

fitlins/cli/run.py
```diff
@@ -60,94 +60,102 @@ def get_parser():
...
+
     warnings.showwarning = _warn_redirect
-    opts = get_parser().parse_args()
-    if opts.debug:
-        logger.setLevel(logging.DEBUG)
+    opts = get_parser().parse_args(args)
+    opts = {k:v for k,v in vars(opts).items() if v is not None} # Remove None
 
-    create_workflow(opts)
+    create_workflow(**opts)
```
effigies: I think I'd rather replace the whole PR with:

```Diff
- def main():
+ def main(args=None):
...
-     opts = get_parser.parse_args()
+     opts = get_parser.parse_args(args)
```

I think this does what you actually want, without doing a lot of refactoring now. (`parse_args` default value for `args` is already `None`, so there's no real need for checking values and branching.)

While there are a couple things I can point to that I think I would probably end up reverting in a refactor, most of my objection boils down to "this is a lot of change in code for a very small change in functionality".

effigies:
And everything is Python 3. I have no intention to support Python 2, though it might still be compatible.

adelavega:
If you what you suggested works, then that works for me for now.

Having said that, of the two options here (allowing `main` to be imported vs allowing `create_workflow` to be imported), I would prefer the latter because it would allow me to use it like a regular python function, rather than having to craft a list of strings as such:

     `main(['/bids/dir', '/output/dir', 'analysis_level', '--nthreads=1'])`

But maybe we can worry about that later and just go with the simple thing for now.

effigies:
Oh, it looked like you were preferring `main` to other targets. So just skipping the shell, but otherwise treating it like a `subprocess.Popen` call rather than a function call. Which made sense to me, because I think `main` will be a more stable API than `create_workflow` for a little while.

So maybe let's go with this for right now. If we do a refactor now, I'm very likely to want to do another once the end goal is clearer in my head, and then we'd have to worry about me breaking your interface when I do go about that.

adelavega:
Yeah that makes sense. I'll test that your solution works and push it soon.

adelavega:
OK, tested it and it seems to work!

effigies:
Ah, I feel like I've seen pathlib, but I haven't taken the trouble to learn it yet.

PR #16, opened 2018-03-27
ENH: Rewrite first level analysis as Nipype workflow

### LINKED ISSUE #12: Rearchitect FitLins as a Nipype workflow
Within a level, each fit/contrast step is independent, making this a good target for parallelization.

This thread is a place to make notes on ideas and steps taken before we get to the stage of opening a PR.

@satra @tyarkoni @adelavega We did a bit of strategizing on this in Austin, but it's taken a bit of a back burner until now. Feel free to add any points from those discussions you think important for directing this, or any ideas you may have had since.

effigies:
This PR tracks progress towards Nipype-ification of FitLins.

I'm starting to settle on an approach that mimics `SpecifyModel` but with a BIDS structure and possible selectors as inputs. The `session_info` objects at present contains sparse and dense components of the design, along with enough metadata to perform file selection (possibly through `BIDSDataGrabber` or a new, similar interface).

It probably makes sense to extend beyond the `SpecifyModel` API further to line up contrasts.

I anticipate the output of this will be shuttled to a `MapNode` to run all first-level estimates and contrasts.

My thoughts on second-level stats are not yet fully fleshed out. I think it would be good to try to head back to pybids and get to the point where we define the design and contrast matrices in such a way as to allow us to run all higher levels in a single shot.

Addresses #12.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -1,0 +1,70 @@
+import nibabel as nb
+from nipype.interfaces.base import (
+    BaseInterfaceInputSpec, TraitedSpec, SimpleInterface,
+    InputMultiPath, File, Directory,
+    traits, isdefined
+    )
+from bids import grabbids as gb, analysis as ba
+
+
+class LoadLevel1BIDSModelInputSpec(BaseInterfaceInputSpec):
+    bids_dirs = InputMultiPath(Directory(exists=True),
+                               mandatory=True,
+                               desc='BIDS dataset root directories')
+    model = File(exists=True, desc='Model filename')
+    selectors = traits.Dict(desc='Limit collected sessions')
+
+
+class LoadLevel1BIDSModelOutputSpec(TraitedSpec):
+    session_info = traits.List(traits.Dict())
+    contrast_info = traits.List(traits.Any())
+
+
+class LoadLevel1BIDSModel(SimpleInterface):
+    def _run_interface(self, runtime):
+        layout = gb.BIDSLayout(self.inputs.bids_dirs)
+        model_fname = self.inputs.model
+        if not isdefined(model_fname):
+            models = layout.get(type='model')
+            if len(models) == 1:
+                model_fname = models[0].filename
+            elif models:
+                raise ValueError("Ambiguous model")
+            else:
+                raise ValueError("No models found")
+
+        selectors = self.inputs.selectors if isdefined(self.inputs.selectors) else {}
+
+        analysis = ba.Analysis(model=model_fname, layout=layout)
+        selectors.update(analysis.model['input'])
+        analysis.setup(**selectors)
+        block = analysis.blocks[0]
+
+        session_info = []
+        for paradigm, _, ents in block.get_design_matrix(block.model['HRF_variables'],
+                                                         mode='sparse'):
+            info = {'entities': ents}
+
+            bold_files = layout.get(type='bold',
+                                    extensions=['.nii', '.nii.gz'],
+                                    **ents)
+            if len(bold_files) != 1:
+                raise ValueError('Too many BOLD files found')
+
+            fname = bold_files[0].filename
+
+            # Required field in seconds
+            TR = layout.get_metadata(fname)['RepetitionTime']
+            # vols = nb.load(fname).shape[3]
+
+            _, confounds, _ = block.get_design_matrix(mode='dense',
+                                                      sampling_rate=1/TR,
+                                                      **ents)[0]
+
+            info['events'] = paradigm
+            info['confounds'] = confounds
+
+            session_info.append(info)
+
+        self._results['session_info'] = session_info
+        return runtime
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -50,7 +50,7 @@ def bids_split_filename(fname):
             fname = fname[:-ext_len]
             break
     else:
-        fname, ext = op.splitext(fname)
+        fname, ext = os.path.splitext(fname)
 
     return pth, fname, ext
 
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -12,6 +12,8 @@ from nipype.interfaces.base import (
 from nipype.interfaces.io import IOBase
 from bids import grabbids as gb, analysis as ba
 
+from ..utils import snake_to_camel
+
 
 def bids_split_filename(fname):
     """Split a filename into parts: path, base filename, and extension
@@ -295,6 +297,8 @@ class BIDSDataSink(IOBase):
                                      self.inputs.in_file):
             ents = {**self.inputs.fixed_entities}
             ents.update(entities)
+
+            ents = {k: snake_to_camel(v) for k, v in ents.items()}
 
             out_fname = os.path.join(
                 base_dir, layout.build_path(ents))
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -5,6 +5,7 @@ from nipype.interfaces.base import (
     InputMultiPath, OutputMultiPath, File, Directory,
     traits, isdefined
     )
+from nipype.interfaces.io import IOBase
 from bids import grabbids as gb, analysis as ba
 
 
@@ -229,9 +230,9 @@ def _copy_or_convert(in_file, out_file):
     if in_ext == out_ext + '.gz' or in_ext + '.gz' == out_ext:
         read_open = GzipFile if in_ext.endswith('.gz') else open
         write_open = GzipFile if out_ext.endswith('.gz') else open
-        with (read_open(in_file, mode='rb') as in_fobj,
-              write_open(out_file, mode='wb') as out_fobj):
-            shutil.copyfileobj(in_fobj, out_fobj)
+        with read_open(in_file, mode='rb') as in_fobj:
+            with write_open(out_file, mode='wb') as out_fobj:
+                shutil.copyfileobj(in_fobj, out_fobj)
         return
 
     # Let nibabel take a shot
@@ -245,15 +246,15 @@ def _copy_or_convert(in_file, out_file):
     raise RuntimeError("Cannot convert {} to {}".format(in_ext, out_ext))
 
 
-class BIDSDataSinkInputSpec(BaseInterfaceSpec):
+class BIDSDataSinkInputSpec(BaseInterfaceInputSpec):
     base_directory = Directory(
         mandatory=True,
         desc='Path to BIDS (or derivatives) root directory')
     in_file = File(exists=True, mandatory=True)
     entities = traits.Dict(usedefault=True,
                            desc='Entitites to include in filename')
-    fixed_entitites = traits.Dict(usedefault=True,
-                                  desc='Entities to include in filename')
+    fixed_entities = traits.Dict(usedefault=True,
+                                 desc='Entities to include in filename')
     path_patterns = InputMultiPath(
         traits.Str, desc='BIDS path patterns describing format of file names')
 
@@ -266,7 +267,7 @@ class BIDSDataSink(IOBase):
     input_spec = BIDSDataSinkInputSpec
     output_spec = BIDSDataSinkOutputSpec
 
-    def _list_ouptputs(self):
+    def _list_outputs(self):
         base_dir = self.inputs.base_directory
 
         layout = gb.BIDSLayout(base_dir)
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -1,7 +1,7 @@
 import os
 from nipype.interfaces.base import (
     BaseInterfaceInputSpec, TraitedSpec, SimpleInterface,
-    InputMultiPath, File, Directory,
+    InputMultiPath, OutputMultiPath, File, Directory,
     traits, isdefined
     )
 from bids import grabbids as gb, analysis as ba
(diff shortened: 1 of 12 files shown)


fitlins/interfaces/bids.py
```diff
@@ -0,0 +1,90 @@
...
+    session_info = traits.List(traits.Dict())
+    contrast_info = traits.List(traits.Any())
+
+
+class LoadLevel1BIDSModel(SimpleInterface):
+    input_spec = LoadLevel1BIDSModelInputSpec
+    output_spec = LoadLevel1BIDSModelOutputSpec
+
+    def _run_interface(self, runtime):
+        layout = gb.BIDSLayout(self.inputs.bids_dirs)
```
effigies: @adelavega @tyarkoni Given your need to specify `Layout` inputs, you may want to chime in on what inputs this `LoadLevel1BIDSModel` interface should have.

effigies:
Alright, I can now create a workflow that gets through the first level models with:

```Python
from nipype.pipeline import engine as pe
from fitlins.interfaces import bids as fib, nistats as fis

bids_dir = '/data/bids/openfmri/ds000030'
preproc_dir = '/data/out/ds000030/derivatives/fmriprep'

loader = pe.Node(
    fib.LoadLevel1BIDSModel(bids_dirs=[bids_dir, preproc_dir]),
    name='loader')
getter = pe.Node(
    fib.BIDSSelect(bids_dirs=preproc_dir,
                   selectors={'type': 'preproc', 'space': 'MNI152NLin2009cAsym'}),
    name='getter')
flm = pe.MapNode(fis.FirstLevelModel(),
                 iterfield=['session_info', 'contrast_info', 'bold_file', 'mask_file'],
                 name='flm')

fitlins_wf = pe.Workflow(name='fitlins_wf', base_dir='/tmp')
fitlins_wf.connect([
    (loader, getter,  [('session_info', 'session_info')]),
    (loader, flm, [('session_info', 'session_info'),
                   ('contrast_info', 'contrast_info')]),
    (getter, flm, [('bold_files', 'bold_file'),
                   ('mask_files', 'mask_file')])
])

fitlins_wf.run(plugin='MultiProc')
```

(`BIDSSelect` may be replaceable with nipype's `BIDSDataGrabber`, but I didn't have the patience to coerce it for a simple but specific case, here. Happy to figure out how to do this, later.)

With something a little more concrete, this might be a good time for people to look at interface specs and raise any issues they might see.

effigies:
Some comments:

* I split `session_info` and `contrast_info` to accommodate other tools that do not perform both the initial parameter estimates and the contrasts in a single interface
* Model loading and data file getting could strictly be done in a single interface, but this approach should allow easy addition of datasets that include useful variables (thinking of Neuroscout here... does this work for y'all?), while restricting file getting to a more confined layout. (Note that variables are selected from `bids_dir` and `preproc_dir`, while files only from `preproc_dir`.)
* "Identity contrasts" (contrast weights to retrieve individual parameter estimates) are generated in the `FirstLevelModel` interface, which duplicates pybids' work. This is because the `get_contrasts()` matrix is by default maximally informative, and depends on the programmer to pare it down as desired. However, this is most easily done with knowledge of the design matrix, which isn't constructed until `FirstLevelModel`. This isn't a problem, but may be a useful data point when thinking about how to evolve the `pybids.analysis` API.

First-level tasks remaining:
* [ ] Data sinking
* [ ] Report generation (some figures are already being generated)
* [ ] Update output specs as second-level needs come into view

Thoughts on second level that are starting to take shape:

* In current FitLins, we sink first-level outputs to a derivatives directory, and then query that directory with a new `BIDSLayout`. This seems somewhat inelegant in a nipype context, in which we should be able to use first-level outputs that don't make it all the way to the derivatives directory. Given that we will be using `MapNode`s, the outputs are necessarily lists corresponding with first-level `session_info` dictionaries, so perhaps the only way to query will be to derive indices from metadata.
* For the sake of minimizing the number of times the `Analysis` object needs to be constructed, it would be nice to be able to collapse all remaining levels into a single design matrix and contrast matrix. It may be worth establishing sooner rather than later whether this is feasible at all and what a timeline would look like, as the complexity of dealing with multiple levels of model in nipype may have a significant cost in time that could be better spent getting this into pybids.

satra:
@effigies - @mgxd is about to run some models, so both getting an understanding of bids model and helping with testing this implementation would be a good interaction.

also, @mgxd can you point to anne's version of the openfmri script. that script has a few things to consider.

for example, because of stimulus correlated motion, or responses (e.g., yes/no, hit/miss) from participants, it's possible that certain contrasts cannot be evaluated in a given run. so if we were to create a run x contrast matrix, some elements can be undefined.

anne's script took this into account when feeding to a first level model, and if a contrast didn't exist for a subject, that subject was discarded for group stats (either all or individually depending on the user). 

i'm not sure how fitlns/bids model plans to deal with missing data, but that may be a consideration to think about.

effigies:
@satra @mgxd Feedback will definitely be appreciated.

As to dealing with missing data, it seems like we should have a way of selecting one approach or another at runtime. It may be getting too close to an implementation detail for BIDS Model (though I am not convinced that it is - it would affect both interpretation and any still-higher-level comparisons) to add that as a parameter in the model spec itself, but unless/until that is decided, a CLI flag shouldn't be too hard here.

It would be good to decide on a default exclusion action; I would probably lean conservatively toward fully excluding a subject with missing data by default. I'm imagining a filter node that will check for missing results in any input field and remove corresponding results from other fields. This could probably be written flexibly to allow different strategies to be implemented in the one node.

tyarkoni:
I'm trying to catch up on this thread. Haven't looked at the code yet, but some quick questions/comments:

* Re: first-level outputs, they don't have to be in a derivatives directory, do they? Can we construct a new `BIDSLayout` from a temporary folder, and use that as the input to the next level? We can use the `merge_layouts` utility to ensure that all of the existing variables etc. are available, if that's an issue. This can be done separately for each subject if need be.

* Re: construction of the `Analysis` object, the expensive step isn't initialization, it's the `setup()` call. But that call accepts any of the standard BIDS keywords, so if this is executed in a `MapNode` context, as long as we pass in the appropriate constraints (e.g., `subject='01'`), I don't see why we should incur any meaningful overhead. If you're talking about the group-level analysis (which I guess is the second level in a canonical summary stats approach, so that makes sense), I'm not sure why we'd need to construct the Analysis more than once. Aren't the mapped nodes collected and piped in together?

* Re: internally duplication of what already happens in `get_contrasts`, is the issue just that you don't want to get back anything but the "identity" contrasts? If so, it would be trivial to add a flag to `get_contrasts` that only returns those contrasts (or rather, we can add `block_contrasts` and `identity_contrasts` that let you independently control the two sets). I think that would be quite useful, actually.

* Re: missing data, I'll have to go back and look, but I recall giving this some thought when working on the `variables` and `analysis` modules in pybids, and I believe missing data should be handled gracefully by default. If I'm not mistaken, you will automatically get a `NaN` populating cells where there are missing values, so it should be easy for estimation software to detect that and make a decision about how to handle such cases. But we can double-check that.

annepark:
@mgxd @satra 

The openfmri script in question can be found here: https://github.com/annepark/openfmri/tree/exclude-copes/subject_level. I've copied an explanation below:

The basic idea was to have the script read a user-defined text file (one for each subject) that shows which copes to combine at the fixed effects level (this was motivated by a project where subjects were missing certain conditions in a variable way). The --exclude_dir argument can be used to point to the directory containing these text files. In the current script, the files need to be named "exclude_copes_[subj_id].txt." Here's an example of how the file would be set up (1 = include, 0 = exclude):

```
cope,run1,run2,run3,run4
1,0,0,0,0
2,0,1,0,1
3,1,1,1,0
```
(the script removes the first row and first column)

For cases like cope1 in the example above, where all runs are being excluded, the script actually just uses all runs, as it normally would. But a text file gets outputted, indicating that this cope should not be used at the group level (this is just a descriptive text file, so the user would have to manually take this into account for group level analyses). Please see the sort_copes() and copes_to_exclude() functions for the major changes to the script.

The create_modelfit_workflow() and create_fixed_effects_flow() workflows were copied directly into the script, in order to make modifications (mainly: changing Nodes to MapNodes so a different fixed effects model could be made for each contrast). 

Note: the get_contrasts() function was also modified (currently commented out), I think as an attempt to automatically change the contrast vector depending on which conditions were missing.

The modifications in their current form are somewhat awkward- happy to provide further explanation if needed!

adelavega:
Just some minor comments:

Should we worry at all that this approach leads to several instances of `BIDSLayout`? I'm guessing the overhead is minimal, but an alternative would be for `LoadLevel1BIDSModel` to also load up the preproc files using a single Layout. IMO, the default place to look for preproc files should be `bidsdir/derivatives/fmriprep` rather than in the output directory. And since, `BIDSLayout` automatically indexes derivatives, if`-p` is `None`, that would work for most scenarios. Unless you think its more clear for users/crash logs if these two Nodes are split up. 

As far as re-using `BIDSDataGrabber`, I suppose you'd want to pass in a single field with all the possible selectors ("session_info"?), rather than specifying various fields? We could amend DataGrabber to optionally take a dictionary of entities, or we you could define one with each entities as its own input like this. 

    bg = BIDSDataGrabber(infields = ['subject', 'run', 'session', 'task', 'type', 'space'], outfields = ['preproc'])
    bg.inputs.base_dir = preproc_dir
    bg.inputs.output_query['preproc'] = {'type': 'preproc', 'space': 'MNI152NLin2009cAsym'}

Regarding additional variables, the way you were doing it previously worked, again because `BIDSLayout` automatically indexes derivatives. So if I just place extra event files in `derivatives`, they are indexed and made available to `Analysis`. What would be useful is the ability to exclude directories, in case there are multiple conflicting event files within `derivatives`.

effigies:
So. Much. Feedback. Thanks, all. I think I'll need to respond in separate comments for aesthetic reasons.

@tyarkoni: 

> * Re: first-level outputs, they don't have to be in a derivatives directory, do they? Can we construct a new `BIDSLayout` from a temporary folder, and use that as the input to the next level? We can use the `merge_layouts` utility to ensure that all of the existing variables etc. are available, if that's an issue. This can be done separately for each subject if need be.

Correct, they don't have to be in a derivatives directory. FitLins is *currently* (i.e. pre-Nipypification) dropping the first level outputs into the derivatives directory and then using a `BIDSLayout` as a way of making sure that the correct first-level statistics are used at the second level.

We could construct a new `BIDSLayout` from a Nipype working directory, but that might be more trouble than just passing entity metadata along to each node that needs to care about it. As I dive more into the second level, I can give a better sense of what kind of querying will be required.

> * Re: construction of the `Analysis` object, the expensive step isn't initialization, it's the `setup()` call. But that call accepts any of the standard BIDS keywords, so if this is executed in a `MapNode` context, as long as we pass in the appropriate constraints (e.g., `subject='01'`), I don't see why we should incur any meaningful overhead.

That's a fair point, and I'll keep that in mind. I don't know if multiple processes making a bunch of disk accesses in parallel will be achieve a reasonable speed-up, due to cache invalidation, but I can give it a shot. While it's expensive enough not to want to repeat unnecessarily, it's not so expensive that running the whole thing once in serial is problematic.

> If you're talking about the group-level analysis (which I guess is the second level in a canonical summary stats approach, so that makes sense), I'm not sure why we'd need to construct the Analysis more than once. Aren't the mapped nodes collected and piped in together?

This will be something of a design decision in terms of the most intuitive way to provide the data to the various GLM interfaces, which are probably always going to be split into levels 1 and 2. My current approach of using a `LoadLevel1BIDSModel` interface is to provide an easily interpretable `session_info` structure, similar to that currently provided by `SpecifyModel`. This could be embedded in a larger list, where each successive element is a higher level, but how to turn a list of unknown (at workflow build time) length into a series of nodes that run each element is unclear at the moment. Also, something like that would pretty much demand another node to select the correct entries to pass to the next level.

Incidentally, if we parallelize `Analysis.setup` at the subject level, I think we'll almost be forced to compute a new `Analysis` object for the group level. I could be wrong on this one.

That said, since constructing an `Analysis` without running `setup()` is cheap, we can at least get the number of blocks cheaply at build time, so maybe this wouldn't be too hard. While minimizing unnecessary computation is a goal, an overriding one here is constructing generally usable interfaces (to be upstreamed into Nipype) that can be intuitively used by users with little exposure to the pybids API and no attachment to FitLins or nistats. I recognize this will increase complexity in some places, but that's the balancing act I'm attempting.

> * Re: internally duplication of what already happens in `get_contrasts`, is the issue just that you don't want to get back anything but the "identity" contrasts? If so, it would be trivial to add a flag to `get_contrasts` that only returns those contrasts (or rather, we can add `block_contrasts` and `identity_contrasts` that let you independently control the two sets). I think that would be quite useful, actually.

Actually, the issue here is again independent use outside the FitLins/nistats context. By my reading, FEAT doesn't need to know about identity contrasts, because it will generate a `pe*.nii` for every column, no matter what.

Further, once I'm in nistats territory, I don't want to depend on pybids in order to construct the full contrast matrix. It's simple enough, anyway. I was highlighting this just to give a data point on `get_contrasts` use. Whether it confirms your approach or inspires you to update the API, I'm equally content. I don't need new functionality at this time.

> * Re: missing data, I'll have to go back and look, but I recall giving this some thought when working on the `variables` and `analysis` modules in pybids, and I believe missing data should be handled gracefully by default. If I'm not mistaken, you will automatically get a `NaN` populating cells where there are missing values, so it should be easy for estimation software to detect that and make a decision about how to handle such cases. But we can double-check that.

Sounds good. We just need to decide what the options for handling such cases are, and which should be considered default in the absence of user/model specification. If the spec defines this, all the better, but FitLins will need to stake out some position, regardless.

effigies:
@adelavega:

> Should we worry at all that this approach leads to several instances of `BIDSLayout`? I'm guessing the overhead is minimal, but an alternative would be for `LoadLevel1BIDSModel` to also load up the preproc files using a single Layout. IMO, the default place to look for preproc files should be `bidsdir/derivatives/fmriprep` rather than in the output directory. And since, `BIDSLayout` automatically indexes derivatives, if`-p` is `None`, that would work for most scenarios. Unless you think its more clear for users/crash logs if these two Nodes are split up. 

I agree that it's probably minimal overhead, which is part of why building in the flexibility doesn't concern me too much. It may be excessive, but it's a little bit going off of experience with the old style of nipype modeling, where you have the `DataGrabber` and `SpecifyModel` interfaces.

The reason I like this is that the model loader doesn't need to know anything about what specific files I need downstream. For example, I could use an iterable to parameterize over `space` and get separate branches in the pipeline for `space-T1w` or `space-fsaverage` (which are GIFTI and will thus need different logic) or whatever. As long as I have the entities to help me pick out the files that correspond to each `session_info`, I can do a lot here.

With regard to defaults, I consider those the concern of the CLI which will build the workflow. The interfaces themselves should be not be over-opinionated, IMO.

> As far as re-using `BIDSDataGrabber` ...

Yeah, I'm not sure quite what I want at this point. I consider it a hack for now, and am unlikely to try to upstream it into nipype. Once the range of uses become clear, I'll think about how to achieve what I want in BDG and what changes will be needed.

> Regarding additional variables, the way you were doing it previously worked, again because `BIDSLayout` automatically indexes derivatives. So if I just place extra event files in `derivatives`, they are indexed and made available to `Analysis`.

That should still be available as is. The thought was just that the file selector and the analysis construction might not always need access the fully merged layout of all inputs.

> What would be useful is the ability to exclude directories, in case there are multiple conflicting event files within `derivatives`.

:+1: That should be simple. I'll add that next.

fitlins/interfaces/bids.py
effigies: Will it be necessary to have both `include` and `exclude`? Or would `inputs.include_pattern` (or `ignore_pattern`, or whatever) do what you need?

I'm fine with putting both in; I just don't want to add options that aren't going to be used.

effigies:
@tyarkoni For `Analysis.setup`, is the IO the majority of the time, or would there be some value to adding an option to only setup the first block (or first `n` blocks)?

tyarkoni:
Not entirely sure, I haven't done any profiling. But most of the IO is the variable loading (i.e., calling `BIDSLayout.get_collections`), and that's done inside `Block.setup`, so if you only set up the first `n` blocks that may well save some time. Then again, in the vast majority of cases, nearly all of the variable information is going to be contained in the 1st level data, so I'm not sure how much it will really save.

Long-term, the optimal way to do this would be to sink the (aggregated) variable data from the 1st level to the same place as the output images, and have subsequent levels read in from there rather than doing it internally in pybids, but we're not there just yet, so for now I think you'll have to set up the first `n` blocks (or just everything).

effigies:
My inclination right now is to go ahead and merge first-level nipypification, so people can start playing with and breaking it while I start work on group/second level.

I've done that currently, with minimal refactoring of the `cli/run.py`. Probably the biggest difference now is that the reports are generated based on the contents of the derivatives directory, as opposed to the specific subprocesses that are run now. The reason for this, following the logic in fmriprep, is that this allows reports to be generated even if some components of the pipeline fail.

The most annoying consequence of this is that explicitly defined contrasts and parameter estimates (identity contrasts) are not distinguishable by file name, at least without reference to the model. They all have the form `*_bold[_space-<space>]_contrast-<contrast>_stat.nii.gz`. This is mostly a matter of reporting, as I would split out parameter estimates and contrasts in the report for the purpose of readability. There isn't much of a statistical distinction to be made.

So given that, we have a couple options:

1) I can break with fmriprep precedent, and use metadata from the workflow to help build the report. This might get a bit messy, with reporting needing to see forward to the `DataSink`'ed filenames, but I'm not sure that will actually be an issue.
2) We can distinguish estimates and contrasts by entity, replacing `contrast-<contrast>` with `estimate-<parameter>`. This proliferates entities a little, which I would be careful of, but it's also probably a pretty common distinction that stats tools will want to draw.
3) In the same vain, following FEAT, which uses `pe/` and `cope/`, we could use `estimate/` and `contrast/` subdirectories. This carries a danger of a high directory-to-file ratio, and becoming more of a navigational hassle than it helps with organization.

Opinions on approach are welcome. @tyarkoni @chrisfilo your opinions on what seems most BIDS-y would be appreciated.

Code reviews are welcome. @adelavega Do you want to check whether I'm breaking anything for you?

Still to-do:

* [ ] Re-constrain participants based on `--participant-label` flag. (This has been lost in the refactor.)

effigies:
With the understanding that reports no longer distinguish between contrasts and estimates, I'm going to merge this. I look forward to the bug reports.

adelavega:
I probably wont be able to test this for a few days as I'm trying to crank some models out using the previous version, but will let you know

satra:
@effigies - are there some existing models i can take a look at?

effigies:
@satra Don't have a good connection at the moment. I'll try to put together a set later tonight or tomorrow morning.

PR #19, opened 2018-04-03
Fixes #18

### LINKED ISSUE #18: Run entity not in contrast path pattern
When outputting contrasts, building a path fails, because one of the entities is 'run' but it is not in the pattern.

https://github.com/poldracklab/fitlins/blob/1e98b93cffbb37091845bbb2ebfac3ed397cb7dc/fitlins/base.py#L164

This is probably due to https://github.com/INCF/pybids/pull/156

Previously, 'run' was erroneously being left out of the entities that came back with the design_matrix, and thus you were not expecting it.

adelavega:
Fixes #18. Pretty simple.

### DIFF
--- a/fitlins/base.py
+++ b/fitlins/base.py
@@ -159,6 +159,7 @@ def first_level(analysis, block, space, deriv_dir):
         for contrast in contrast_matrix:
             stat_ents = preproc_ents.copy()
             stat_ents.pop('modality', None)
+            stat_ents.pop('run', None)
             stat_ents.update({'contrast': snake_to_camel(contrast),
                               'type': 'stat'})
             stat_fname = op.join(deriv_dir,


effigies:
Sorry if this is a dumb question, but are you expecting runs to be collapsed in the first level? I'd assume you'd do each run separately, then combine results at higher levels.

In which case maybe the better approach would be to add `[run-<run>_]` (or similar) to the patterns.

adelavega:
Oh, well I guess its up to you. I was wondering about that and just assumed they were collapsed at the first level because of the path pattern you were using. When are the first level runs combined as it currently stands then?

effigies:
Nah, it was an oversight because all of the datasets I was testing with only had one run of the task I had a model for.

I would assume where they're combined is defined by the model, but if there should be some implicit behavior here, we'll need to decide what it is. I'll check whether the spec says anything in particular.

adelavega:
Ah cool. Well we can probably punt on this for a second, it seems non trivial to allow flexible specification of how to combine runs. Although I would expect the default or modal approach to be an FSL style approach with a fixed effects level in between.

effigies:
Closed by #20.

PR #20, opened 2018-04-03
ENH: Add acq, rec, run and echo to output patterns

### LINKED ISSUE #18: Run entity not in contrast path pattern

effigies:
The original pattern for functional images is:

```
sub-{subject}[/ses-{session}]/func/
    sub-{subject}[_ses-{session}]_task-{task}[_acq-{acquisition}][_rec-{reconstruction}][_run-{run}][_echo-{echo}]_{type<bold>}.nii.gz
```

Any of these may appear in preprocessed files, so we should be prepared to distinguish between their respective outputs.

Fixes #18.

### DIFF
--- a/fitlins/base.py
+++ b/fitlins/base.py
@@ -27,14 +27,20 @@ plt.rcParams['svg.fonttype'] = 'none'
 plt.rcParams['image.interpolation'] = 'nearest'
 
 PATH_PATTERNS = (
-    '[sub-{subject}/][ses-{session}/][sub-{subject}_][ses-{session}_]'
-    'task-{task}_bold[_space-{space}]_contrast-{contrast}_{type<stat>}.nii.gz',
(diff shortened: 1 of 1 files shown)


effigies:
@adelavega Do you want to test on this branch for a bit before we merge? I don't think there's likely to be any harm, so if it's easier to test off of master, I'm fine with that, too.

adelavega:
Sure, I can do that right now

PR #22, opened 2018-04-04
Create group level results dir - fix #21

### LINKED ISSUE #21: Group level folder not created
When creating the group level svg plot the following error occurs:

    FileNotFoundError: [Errno 2] No such file or directory: '/analysis/out_all2/fitlins/ses-
    perceptionTraining01/ses-perceptionTraining01_task-perception_run-1_bold_contrasts.svg'

Looks like the parent folder is never created prior to outputting the SVG, hence the error.

adelavega:
Fix to #21 by creating the parent folder.

There may be a more elegant way to do this rather than just assuming that will be the first time the folder is needed. But for the time being this fixes the bug.

### DIFF
--- a/fitlins/base.py
+++ b/fitlins/base.py
@@ -244,6 +244,8 @@ def second_level(analysis, block, space, deriv_dir):
             deriv_dir,
             fl_layout.build_path(contrasts_ents, strict=True))
 
+        # Make parent results directory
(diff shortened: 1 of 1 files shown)


effigies:
Left a bit of a philosophical note on #21, but this looks like the right thing to do, no matter what the end decision on pruning entities is.

PR #23, opened 2018-04-05
WIP: Pybids updates, multiple runs

adelavega:
First, some pybids related updates on a PR i'm working on there:
- [x] Pybids Analysis no longer needs selectors to be fed in explicilty 
- [x] get_contrasts should only return identity contrasts for 'variables' in model (e.g. not extraneous ones)

And some additional changes
- [x] Suggested way of separating CLI logic (sys.exit) from argument parsing, to make it easier to use as a library. Sys.exit was previously making it difficult to debug with ipdb
- [x] Ensure models w/ multiple runs work


adelavega:
I believe the last commit I made fixes an issue which prevent models with 2+ runs, 2+ subjects, and 2+ contrasts from running. The `data` variable was used in two different ways (I'm not sure though, please take a look).

Now it looks like it working, and I'm getting sensible looking results from these models.

PR #26, opened 2018-04-30
ENH: Run auto model by default

### LINKED ISSUE #24: When no model found, use default model
Automatically-generated models were introduced into pybids with incf/pybids#119. In the absence of a model, we should run an automatically generated model.

effigies:
Closes #24

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -61,7 +61,8 @@ class LoadLevel1BIDSModelInputSpec(BaseInterfaceInputSpec):
     bids_dirs = InputMultiPath(Directory(exists=True),
                                mandatory=True,
                                desc='BIDS dataset root directories')
-    model = File(exists=True, desc='Model filename')
+    model = traits.Either('default', File(exists=True),
+                          desc='Model filename')
     selectors = traits.Dict(desc='Limit collected sessions')
     include_pattern = InputMultiPath(
         traits.Str, xor=['exclude_pattern'],
@@ -99,6 +100,8 @@ class LoadLevel1BIDSModel(SimpleInterface):
                 raise ValueError("Ambiguous model")
             else:
                 raise ValueError("No models found")
+        elif model_fname == 'default':
+            model_fname = ba.auto_model(layout)
 
         selectors = (self.inputs.selectors
                      if isdefined(self.inputs.selectors) else {})


effigies:
The interface I set up assumes one model instead of a list of models, and they're loaded in-interface. If we can't combine it into one big model, I'll need to figure out the best way to handle lists.


-------- Original message --------

effigies:
Well, the point here would be to auto run, so I wouldn't assume enough selectors from users to get a unique model, as currently built. If we can't have a single model that handles all the tasks, then I should handle lists over here.

tyarkoni:
I think if you're going to go down that road (which seems reasonable), you'll need to implicitly loop over all detected tasks and run the whole fitlins pipeline for each one... so it seems like a more general problem than what's happening in `auto_model`.

effigies:
I was thinking of combining in the sense of building a model as a series of independent chains (i.e. block n might depend on blocks 1..n-1, but doesn't need to depend on all). The goal would not be to change the semantics; just the organization.

tyarkoni:
@Shotgunosine I think in principle the approach you suggest already works under the current spec. But fitting one big model across multiple tasks is something of an edge case, and has a different meaning (and will produce different results) from the typical scenario, where one is more likely to have several different models, each of which needs to be fit to a different task.

@effigies I may be misunderstanding what you're saying, but the models coming back from `auto_model` always only apply to `level='run'`. The list of models you get back would presumably have to be run completely in parallel; I don't think there can be any dependency between the elements of that list.

effigies:
@tyarkoni I agree each element of the list should be independent. I was suggesting there might be (I don't know) a way of combining into a single model where the blocks are independent.

effigies:
Just pushed a proposed refactoring, which always assumes multiple independent models. I suspect flattening inside the node will probably end up producing an untenable situation when we try to recombine at the second level... It's possible that what we want is to just load the models (whether they're json files or auto-generated) and construct entirely independent pipelines for each.

tyarkoni:
I can't think of any situation in which a user would want/need to recombine models at the second level in a way that can be easily automated. I think we can safely construct completely independent pipelines for each task (beginning with the subsetting of the input image list based on the `task`).

effigies:
Think this is ready, by the way. Reviews welcome. Will merge tomorrow unless somebody requests more time to review.

PR #28, opened 2018-05-10
Neuroscout changes / pybids updates

adelavega:
WIP

Various minor changes I've had to make to get Neuroscout models running. So far, all changes relate to changes to pybids/grabbit.


adelavega:
Okay, there is now a problem with `BIDSDataSink`. It seems @effigies was using `BIDSLayout` on an empty folder ("fitlins" derivatives) to generate paths for outputs. However, it is no longer valid to initiate a Layout on an empty folder because a `dataset_description.json` file is required. In other words, it is not the intended behavior to instantiate a `BIDSLayout` on anything but a BIDS-compliant dataset. I tried using only a `derivatives` config file, but that threw another error (@tyarkoni, this should be allow right? If so, I'll open an issue in pybids)

Not sure how to proceed here, because it seems like a common use case to use `BIDSLayout` like this. Maybe it's best to allow an argument to change the root for generated paths? And point the Layout to the original dataset?

effigies:
We should just auto-generate a `dataset_description.json`. I think that's a good move for BIDS apps in general.

adelavega:
Hmm...  that would work, but I don't see an output folder as a "dataset", and as is, pybids only requires "bids" not "derivatives" configs. It's really a "derivative". Maybe what would work here is to initalize the BIDSLayout on the main dataset, but add the output folder as a "domain", and have functionality to generate paths for the sub-domain, relative to its root. 

(What is unclear to me is if its valid for the root layout dir to not be a "bids" config, hence my note to Tal above. The error it throws is related to the fact that pybids expects all the getters that are populated by having a "bids" config as the root config).

fitlins/base.py
```diff
@@ -152,13 +154,14 @@ def second_level(analysis, block, space, deriv_dir):
             paradigm = pd.DataFrame(cols)
 
             fmri_glm.fit(data, design_matrix=paradigm)
-            stat_type = [c['type'] for c in block.contrasts if c['name'] == contrast][0]
+            stat_type = [c['type'] for c in block.contrasts if c['name'] == contrast] or ['T']
+            stat_type = stat_type[0]
```
effigies: What's the case that's being considered here? Isn't `'type'` required?

fitlins/cli/run.py
```diff
@@ -77,8 +77,8 @@ def get_parser():
                              'removed)')
     g_bids.add_argument('-m', '--model', action='store', default='model.json',
                         help='location of BIDS model description (default bids_dir/model.json)')
-    g_bids.add_argument('-p', '--preproc-dir', action='store', default='fmriprep',
-                        help='location of preprocessed data (default output_dir/fmriprep)')
+    g_bids.add_argument('-p', '--preproc-dir', action='store', default=None,
+                        help='location of preprocessed data (default bids_dir/fmriprep)')
```
effigies: `bids_dir/derivatives/fmriprep`?

I take it there's a case where you want your outputs in a different derivatives directory?

fitlins/interfaces/bids.py
```diff
@@ -113,9 +113,11 @@ def _run_interface(self, runtime):
...
 
 class LoadLevel1BIDSModelInputSpec(BaseInterfaceInputSpec):
-    bids_dirs = InputMultiPath(Directory(exists=True),
-                               mandatory=True,
-                               desc='BIDS dataset root directories')
+    bids_dir = Directory(exists=True,
+                         mandatory=True,
+                         desc='BIDS dataset root directory')
+    preproc_dir = Directory(exists=True,
+                            desc='Optional preprocessed files directory')
```
effigies: Maybe:

```Python
confounds_dir = InputMultiPath(Directory(exists=True),
                               desc='Derivatives directories containing confounds files')
```

fitlins/interfaces/bids.py
```diff
@@ -250,20 +263,21 @@ def _run_interface(self, runtime):
...
                               "\n\t".join(
                                   '{} ({})'.format(
                                       f.filename,
                                       layout.files[f.filename].entities)
                                   for f in bold_file)))
 
             # Select exactly matching mask file (may be over-cautious)
-            bold_ents = layout.parse_file_entities(bold_file[0].filename)
+            bold_ents = layout.parse_file_entities(bold_file[0].filename,
+                                                   domains=['bids'])
```
effigies: What does this `domains` do? I'm guessing this will cause problems with `space` entities.

tyarkoni:
Have not looked at the code in detail, but re: @adelavega's suggestions about, that sounds like the way I would do it. What path issues are you having for the `derivatives/` domain?

fitlins/base.py
```diff
@@ -66,9 +66,12 @@ def expand_contrast_matrix(contrast_matrix, design_matrix):
 
 
 def init(model_fname, bids_dir, preproc_dir):
-    analysis = ba.Analysis(model=model_fname,
-                           layout=grabbids.BIDSLayout([bids_dir, preproc_dir]))
-    analysis.setup(**analysis.model['input'])
+    layout = grabbids.BIDSLayout(
+        bids_dir,
+        config=[('bids', bids_dir), ('preproc', preproc_dir)])
```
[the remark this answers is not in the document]
adelavega: good catch

fitlins/base.py
```diff
@@ -83,9 +86,8 @@ def second_level(analysis, block, space, deriv_dir):
     analyses = []
 
     # pybids likes to give us a lot of extraneous columns
-    cnames = [contrast['name'] for contrast in block.contrasts]
     fmri_glm = level2.SecondLevelModel()
-    for contrasts, idx, ents in block.get_contrasts(names=cnames):
+    for contrasts, idx, ents in block.get_contrasts():
```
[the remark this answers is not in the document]
adelavega: Yes, previously the "variables" field was not doing anything. Now it will apply a "select" operation, so only the variables you care about are kept, and only contrasts for those variables will be produced here.

fitlins/base.py
adelavega: I'm trying to rack my brain, but for some reason I recall it wasn't? and T should be default? But maybe I'm wrong, so I'll check the spec.

fitlins/cli/run.py
adelavega: No, I've mentioned this before, but `bids_dir/derivatives/fmriprep` is a much more sensible default for me. That's where I would think 90% of people would place their fmriprep directory. Doesn't make sense for it to be in the "output" folder for me (especially if people are not putting the fitlins outputs in derivatives)

But maybe the description should be changed. if preproc_dir = None, and only ` bids_dir` is passed to `BIDSLayout`, all derivatives will be recursively indexed. So if there's an fmriprep folder in there, the preproc files will be available by default.

fitlins/interfaces/bids.py
adelavega: Grabbit now implements "domains" when it indexes. The default ones are 'bids' and 'derivatives', but it will add any other folders with a `layout.json` as a `domain`.

And yes I think you're correct about the `space` entity, but I was trying to workaround a grabbit bug. When fixed, I think we shouldn't have to specify a domain, and it should collapse across all.

adelavega:
I'm not able to replicate the error in a different env, and I think @effigies is correct that a dataset_description should be generated, so I think the following is the way to deal with it, to make sure all the needed entities are available:

    layout = BIDSLayout('/output_dir/', config=['bids', 'derivatives'])

I'll generate a very simple `dataset_description.json` for now, and the details of the file can be iterated later.

fitlins/cli/run.py
effigies: The output directory is a derivatives directory, kind of by definition, since we put the fitlins derivatives in it. Similarly, with fmriprep, that's where we look for FreeSurfer derivatives. It seems much more common to me to keep your derivatives separate from your raw data than to have some of your derivatives in `bids_root/derivatives` and some out of it.

But this may be a false impression. What's your use case where you aren't putting your fitlins outputs next to you fmriprep outputs?

effigies:
Is there any validation on `dataset_description.json`? The fields for BIDS proper and BIDS derivatives are fairly different. Should all BIDS fields be valid derivatives fields? (@chrisfilo you might have more context on original intent than I.)

chrisfilo:
Yes - the validator currently uses this schema: https://github.com/INCF/bids-validator/blob/master/validators/schemas/dataset_description.json

effigies:
I meant in terms of derivatives `dataset_description.json` files being a superset of BIDS `dataset_description.json` or a separate thing entirely. e.g. `Name` and `BIDSVersion` are not mentioned in BEP003.

chrisfilo:
I don't think any validation of any part of derivatives extensions have been implemented yet if this is what you are asking...

As for the standard we should keep them in sync. Ideally derivatives one only adding some fields.

adelavega:
Sounds good. Those are the only two fields required by pybids right now on init.

fitlins/interfaces/bids.py
effigies: Unless you've already worked on this, let's leave it for another PR. This gets where we need to be in separating the directories.

fitlins/cli/run.py
effigies: I'm okay with leaving this as you have it for right now. Before the next release we should decide whether your approach or mine is the long-term one. I don't want to change CLI defaults too often.

effigies:
Okay, with #31, and the fixes already agreed upon, I think this can go in. Then I'll check on incf/pybids#176.

fitlins/cli/run.py
adelavega: My derivatives are going in `bids_root/derivatives` but a bit deeper than that. E.g.: `bids_root/derivatives/neuroscout/analysis_number/`

It was my impression that most ppl would keep all derivatives w/ their raw data, but maybe not... worth polling and asking around a bit. 

I could imagine ppl might treat `fitlins` outputs a bit differently than your average derivative output since it will be run many times, for many models. Whereas you might only run fmriprep and freesurfer once.

adelavega:
The explain the last commit: `Analysis.setup()` now defaults to using `model['inputs']` for `setup`. if any kwargs are provided, the inputs are updated (dictionary update).

The current code was causing problems when no participants were specified, but the inputs were more restrictive, because setup for all was requested.

fitlins/workflows/base.py
```diff
@@ -5,29 +5,33 @@
...
     if not all_models:
         raise RuntimeError("Unable to find or construct models")
 
+    selectors = {'subject': participants} if participants is not None else {}
+
     loader = pe.Node(
-        LoadLevel1BIDSModel(bids_dirs=[bids_dir, preproc_dir],
-                            selectors={'subject': participants}),
+        LoadLevel1BIDSModel(bids_dir=bids_dir,
+                            preproc_dir=preproc_dir,
```
effigies: You can't set `preproc_dir=None`, which is the default. This will need to be:

```Python
loader = pe.Node(
    LoadLevel1BIDSModel(bids_dir=bids_dir, selectors=selectors),
    name='loader')
if preproc_dir is not None:
    loader.inputs.preproc_dir = preproc_dir
```

effigies:
This line is now failing:

https://github.com/poldracklab/fitlins/blob/8cda5a62d9d65c17f5ca900fd80a6c4f0025613b/fitlins/interfaces/bids.py#L188-L190

```
  File "/src/fitlins/fitlins/interfaces/bids.py", line 188, in _run_interface
    names = [col for col in confounds.columns
AttributeError: 'NoneType' object has no attribute 'columns'
```

So it looks like `get_design_matrix` is returning `None` now:

https://github.com/poldracklab/fitlins/blob/8cda5a62d9d65c17f5ca900fd80a6c4f0025613b/fitlins/interfaces/bids.py#L180-L182

fitlins/workflows/base.py
adelavega: Ah. Is there no way to make it nullable with Traits?

effigies: You could define a trait like that, but I'd rather do type wizardry in Nipype than here.

adelavega:
Weird, not having the same issue. Any more insight into why? I also don't believe I changed that, so it could be related to pybids. Do you have the latest github version?

effigies:
I'm using your branch that's supposed to fix the selectors issue.

effigies:
Hmm. Actually, that may not be true. I'm now getting:

```
  File "/src/fitlins/fitlins/interfaces/bids.py", line 159, in _run_interface
    analysis.setup(**selectors)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/analysis.py", line 89, in setup
    b.setup(input_nodes, identity_contrasts, **selectors)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/analysis.py", line 200, in setup
    transform.select(coll, model['variables'])
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 87, in __new__
    return t.transform()
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 223, in transform
    self._clone_variables()
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 123, in _clone_variables
    for v in self.variables}
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 123, in <dictcomp>
    for v in self.variables}
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/variables/kollekshuns.py", line 176, in __getitem__
    return self.variables[var]
KeyError: 'FramewiseDisplacement'
```

So I don't even get to that point, any more. I'm not sure what combination of versions I had to get that far.

Do we know what the state of pybids/grabbit are? Are they known to be broken, and there's a roadmap back to working, or is this working, and I just need to learn how to do things in the refactored world?

adelavega:
Pybids / grabbids should not be broken, AFAIK. Pybids has not released in a while though, but it looks like you're running the latest version off my PR over there. My guess here is that this has to do with the Layout. Is your preproc folder nested in your bids folder? If not, that may be the culpirt. My guess is this may relate to: https://github.com/INCF/pybids/issues/179

effigies:
> Is your preproc folder nested within the bids_dir?

No.

> If not, that could be the culprit, since it would be a different domain, and its not getting read by get_events.

Okay, so what are domains, and what purpose do they serve? And is there any way I can get back to pulling in all variables that I need without manipulating the filesystem?

tyarkoni:
Domains are basically rulesets. They don't have to have separate search paths/roots (though they often will); the main reason they exist in the BIDS context is so that we can apply derivatives extensions specs separately from core BIDS rules (potentially to different folders).

effigies:
Is there any way to get `get_events` to load confounds if they're not in `<bids>/derivatives/fmriprep`?

tyarkoni:
Certainly should be... if it's not possible, that's a bug. Sorry, I haven't had a chance to catch up on the various pybids-related threads, but will try to do that today.

tyarkoni:
Actually, I guess it's not clear whether `get_events` is expected to retrieve anything other than information that was stored in `events.tsv` files. As currently implemented, `get_events` only deals with information stored in `*events.tsv` files. I would prefer to actually deprecate `get_events` at some point in favor of `get_collections`, which returns any/all available variables. Can you get confounds out of `get_collections` at the moment?

effigies:
I don't think that works:

```Python
>>> layout.get_collections('run', types=['confounds'])
[]
```

But I have no idea how `get_collections` is supposed to work. I've only ever used it by calling `analysis.setup()`.

adelavega:
Yeah sorry I meant to say, if you have your folder *not* nested, it won't find the events/confounds because of a bug I have also been experiencing. I'll fix it.

adelavega:
Actually, the bug is clearly more fundamental, since `Analysis` is using `get_collections` not `get_events` for confounds. `get_events` should only get/merge event files.

tyarkoni:
Can somebody summarize for me--either here or on slack--what the outstanding issues are? And is there some code I can run to reproduce the issues?

effigies:
New error.

#### CLI

```
docker run --rm -it -v /data/scratch/ds000030:/scratch \
    -v /data/bids/openfmri/ds000030:/data:ro \
    -v /data/out/ds000030/derivatives:/out \
    -v /data/cjmarkie/Projects/crn/fitlins/fitlins:/src/fitlins/fitlins \
    poldracklab/fitlins:neuroscout /data /out participant -p /out/fmriprep -w /scratch
```

#### Traceback

```
  File "/src/fitlins/fitlins/interfaces/bids.py", line 160, in _run_interface
    analysis.setup(**selectors)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/analysis.py", line 89, in setup
    b.setup(input_nodes, identity_contrasts, **selectors)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/analysis.py", line 202, in setup
    coll = apply_transformations(coll, self.transformations)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/analysis.py", line 470, in apply_transformations
    func(collection, cols, **kwargs)
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 87, in __new__
    return t.transform()
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 223, in transform
    self._clone_variables()
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 123, in _clone_variables
    for v in self.variables}
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/analysis/transformations/base.py", line 123, in <dictcomp>
    for v in self.variables}
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/bids/variables/kollekshuns.py", line 176, in __getitem__
    return self.variables[var]
KeyError: 'trial_type'
```

PR #29, opened 2018-05-11
ENH: Make dataset description

effigies:
Following BEP 003:

> ## Derived dataset and pipeline description
> `<dataset>/derivatives/<pipeline_name>/dataset_description.json`
> 
> Keys:
> * HowToAcknowledge
> * PipelineDescription
>   * Name
>   * Version
>   * CodeURL
>   * DockerHubContainerTag
>   * SingularityContainerURL
>   * SingularityContainerMD5
> * SourceDatasetsURLs - a list of URLs to the source dataset(s) (mandatory)
> * SourceDatasetsVersions - a list of versions of the source dataset(s) (optional only if SourceDatasetsURLs specify the version unequivocally)
> * License

We should start adding docker hub tags and singularity URLs as environment variables when we build images. I've put in a stub to get the MD5, which will attempt to pull the `'version'` tag from Singularity Hub.

BIDS doesn't specify a field for version, which seems like an oversight, but perhaps there's another mechanism for marking it?

Long-term, it would be good to standardize the construction of this metadata in the BIDS-App spec and pybids.

Related to discussion in #28.

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -135,9 +135,7 @@ def create_workflow(opts):
     deriv_dir = op.join(output_dir, 'fitlins')
     os.makedirs(deriv_dir, exist_ok=True)
 
-    desc = op.join(deriv_dir, 'dataset_description.json')
-    with open(desc, 'w') as fobj:
-        json.dump({'Name': 'FitLins output', 'BIDSVersion': '1.1.0'}, fobj)
+    bids.write_derivative_description(bids_dir, deriv_dir)
 
     # BIDS-Apps prefers 'participant', BIDS-Model prefers 'subject'
     level = 'subject' if opts.analysis_level == 'participant' else opts.analysis_level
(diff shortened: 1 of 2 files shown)


adelavega:
Okay, I think you're right about this. I was mostly confused about the semantics of calling this a "dataset", but I see that such a file was recommended in the original derivatives BEP.

fitlins/cli/run.py
```diff
@@ -135,9 +135,7 @@ def create_workflow(opts):
     deriv_dir = op.join(output_dir, 'fitlins')
     os.makedirs(deriv_dir, exist_ok=True)
```
adelavega: You can omit this line, since the dir is also made in `write_derivative_description`

effigies: Does it make more sense for `write_derivative_description` to assume that `deriv_dir` exists?

adelavega: Probably, although that's what's nice was `exist_ok`, is you can use it either way.

adelavega: If you think in the future there may be other files you might want to instantiate in the folder, you could call the function something like `create_derivative_template` or something

effigies: Okay. For now I'm just going to let `create_workflow` do the `makedirs`. Might change again in the future.

PR #30, opened 2018-05-11
RF: Second-level workflow

effigies:
Going to start this PR before it's anywhere close to ready, because I'm having trouble thinking through exactly what the second level should look like.

`LoadBIDSModel` contrast matrices and entity dictionaries are now output as lists of lists, where the element corresponds to the model level. `LoadBIDSModel().contrast_info[0]` corresponds to the earlier `LoadLevel1BIDSModel().contrast_info`, and similar for `entities`.

For the first level, the contrast matrix (transposed and with a `type` column added) looks like the following:

  | accept | explode | type
-- | -- | -- | --
accept_vs_explode | 1 | -1 | T

The columns here correspond to variables that are calculated in the design matrix. For the second level, the columns correspond to first level statistical maps, and actually all have the same name:

  | accept_vs_explode | accept_vs_explode | ... | type
-- | -- | -- | -- | --
group_accept_vs_explode | 1 | 1 | ... | T

This does result in issues with HDF5 that I'm currently solving by replacing the column names with sequential integers, which I suspect is going to cause problems if we ever have a model with multiple contrasts at the second level. Also, if we ever want to use the implicit contrasts (parameter estimates) at the second level.

Does it make sense for pybids to transform names to somehow include the entities associated with the input maps? That will probably also make selecting the correct input files easier.

Anyway, I think I'm going to need two new interfaces here:

1) BIDS: Correctly group first-level outputs into the inputs for each second-level node
2) Nistats: Run second-level stats

The first is going to be the complicated bit, but the second will require it to be done correctly.

One thing that would be useful would be a three-level model with multiple contrasts at the first level and multiple at at least one of the higher levels. This should help with testing re-grouping node edge cases.


tyarkoni:
We could in principle transform names in pybids, but it's not immediately obvious to me how much that would help... I assume that in most cases, the second-level entities will be pretty much identical for all contrasts, since they'll typically be from the same `task`, and all the other entities (`subjects`, `runs`, etc.) will have dropped out.

I guess the way I think about propagation across levels is that the concatenation typically happens along rows rather than columns. The way pybids currently structures things internally is like this:


| name              | subject |
|-------------------|---------|
| explode           | 1       |
| explode           | 2       |
| accept            | 1       |
| accept            | 2       |
| accept_vs_explode | 1       |
| accept_vs_explode | 2       |

I'm not sure you need to be storing the contrast information from the 1st level at the 2nd level, which may be where you're running into issues? AFAICT the 2nd level analysis doesn't need to know or care how the `accept_vs_explode` contrast was generated. It just knows that there are N 1st-level maps called `accept_vs_explode` that you want to do a t-test on. Any custom weights that need to be applied (other than just 1 vs. 0 contrasts propagated forward) have to be specified at the current level anyway. So I think you only need to store the names, entities, and image data for contrast images at the previous level, and not their prior contrast information or history.

Does that help, or am I misunderstanding the problem?

effigies:
> I guess the way I think about propagation across levels is that the concatenation typically happens along rows rather than columns. The way pybids currently structures things internally at the second level is like this:
> 
> | name              | subject |
> |-------------------|---------|
> | explode           | 1       |
> | explode           | 2       |
> | accept            | 1       |
> | accept            | 2       |
> | accept_vs_explode | 1       |
> | accept_vs_explode | 2       |

Okay, but this subject information isn't available in `get_contrasts`. If it's in `get_design_matrix`, that may help. I've been ignoring it, since it was easier to just parse the output directory, previously.

Anyway, the reason things are in columns instead of rows is that Pandas complains if you try to put a `'T'`/`'F'` value at the bottom of a numeric column, which is how I get the contrasts to nistats.

> I'm not sure you need to be storing the contrast information from the 1st level at the 2nd level, which may be where you're running into issues? AFAICT the 2nd level analysis doesn't need to know or care how the `accept_vs_explode` contrast was generated.

Right, but I need to care enough to get it the maps corresponding to that contrast.

> So I think you only need to store the names, entities, and image data for contrast images at the previous level, and not their prior contrast information or history.

I'm not sure what you mean by name here. The contrast name?

Also, is there any chance that someone will want the identity contrasts at a higher level, because having to recombine the parameter estimates and the contrasts is going to be a pain. I'll probably just have to collapse them all into contrasts, and then find some post-hoc way of separating them out again.

tyarkoni:
> Okay, but this subject information isn't available in get_contrasts. If it's in get_design_matrix, that may help. I've been ignoring it, since it was easier to just parse the output directory, previously.

Entities should be returned from `get_contrasts`... the `ContrastInfo` namedtuple has a `.entities` field. Are you not seeing that at all, or is it coming back, but without `subject`?

> Anyway, the reason things are in columns instead of rows is that Pandas complains if you try to put a 'T'/'F' value at the bottom of a numeric column, which is how I get the contrasts to nistats.

If the issue is just HDF5 complaining, maybe store it as above, and then transpose before passing to pandas?

> Right, but I need to care enough to get it the maps corresponding to that contrast.

I guess I don't see why you need information about the previous contrast weights to do that... shouldn't the contrast name plus subject be sufficient?

> I'm not sure what you mean by name here. The contrast name?

Yes.

> Also, is there any chance that someone will want the identity contrasts at a higher level, because having to recombine the parameter estimates and the contrasts is going to be a pain. I'll probably just have to collapse them all into contrasts, and then find some post-hoc way of separating them out again.

Not sure I follow. What kind of recombination do you have in mind? The general principle is that the user should explicitly forward any contrast maps the expect to need at a later level; there is no provision (and I don't think there should be) for them to reach back several levels and pull something out that they're not explicitly propagating. So, e.g., if you define `accept`, `explode`, and `accept_vs_explode` at the first level (run), and then only define a `t` contrast on `accept_vs_explode` at the 2nd level (subject), you will end up with only one set of maps (`accept_vs_explode`) at the 3rd level (group). At that point you cannot (and should not expect to) define a new contrast that weights `accept` and `explode` differently. If you need to do that, you can simply turn `auto_contrasts` on at the 2nd level to bring all 3 sets of maps forward to the 3rd level.

effigies:
Sorry for the delay. I feel like we're talking past each other, so probably I wasn't very clear in framing this. Possibly I don't know what the problems are at this point. I'm mostly having trouble thinking through how to actually achieve the things that I think we both agree are needed.

tyarkoni:
Sorry about the delay addressing all of these issues; I'm teaching a short course this week that's consuming all my time, but that ends this afternoon, so I'll get back to this shortly.

effigies:
Ohhhh wow. I just finally saw the "index" entry in second-level contrasts, and there are my missing entities. Okay. I think I see where I'm going with this.

effigies:
Breaking things.

PR #38, opened 2018-05-22
RF: Simplify entry-points, restore preproc discovery

### LINKED ISSUE #32: RFC: Default location for preproc directory
Given the BIDS App CLI:

```
$ app <bids> <out> <level> [<option> ...]
```

We should settle on a default location to look for the preprocessed dataset. In the case of:

```
$ fitlins /data/bids/ds000XYZ /data/bids/ds000xyz/derivatives <level> [...]
```

The reasonable default is `/data/bids/ds000xyz/derivatives/fmriprep`, but this is `derivatives/fmriprep`, relative to the BIDS root, and `fmriprep` relative to the output directory. When we have:

```
$ fitlins /data/bids/ds000XYZ /data/out/ds000xyz/derivatives <level> [...]
```

The options become:

* `/data/bids/ds000xyz/derivatives/fmriprep`
* `/data/out/ds000xyz/derivatives/fmriprep`

Up until now I have assumed the latter, reasoning that derivatives will tend to be kept together, even when not falling under the BIDS root directory. However, in #28, the proposed presumption is that preprocessed derivatives are special, and will be kept in the BIDS root directory even when the Fitlins derivatives to be written will go elsewhere.

It would be nice to get some intuitions from others. (I'll go ahead and tag @tyarkoni, @satra, @mgxd, @jordandekraker.)

To preempt what seems like an inevitable suggestion, I'm hesitant to make this *too* smart, where we first check for one and then the other, as adding the higher priority one to a setup that was using the lower-priority would silently change the behavior (assuming that the preprocessed files are different). A case to consider here is if the higher-priority one was misspelled (e.g. `frmiprep`), and we will silently fall back to the lower-priority one, correctly spelled directory. But maybe that's okay?

Finally, just as a UX note, the current behavior allows relative paths as well as absolute. So the actual default is `-p fmriprep`, which is resolved relative to `out_dir`. The idea here is to allow alternatives like `-p spm` or `-p feat` without requiring users to add a full path. So making the right default will affect ease-of-use in that case.

### LINKED ISSUE #34: Abstract CLI-specific logic out
To make it easier for me (or others) to use fitlins as a "library", it would be useful to abstract out logic specific to the CLI.

However, I know its not the intention for fitlins to be generally used a library, so I propose some minimalist changes. Currently the only issue I'm having relates to the `sys.exit` and retcode. Since `fitlins` forces a `sys.exit` this will also exit Neuroscout, and prevent me from doing any other operations following execution. An easy fix would be to add a `run` function that does what `main` does now (parse args, run `create_workflow`, etc), but simply returns the retcode. `main` would then call `run` and execute `sys.exit`. I would then import `run` instead of `main. 

Alternatively, parse args logic could be moved to `create_workflow` and `sys.exit` could be called in `main`, and I would import `create_workflow`.

effigies:
This places `sys.exit` in the `main` entry-point, to reduce the chance of unexpected exits for any external calling code.

This also restores preproc discovery, prioritizing that found in `<bids_dir>/derivatives`.

Other minor simplifications were also applied, such as allowing `argparse` to apply `os.path.abspath` to file inputs.

Closes #32 
Closes #34

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # pull request on our GitHub repository:
 #     https://github.com/kaczmarj/neurodocker
 #
-# Timestamp: 2018-04-18 20:02:44
+# Timestamp: 2018-05-22 20:52:23
 
 FROM neurodebian@sha256:5fbbad8c68525b588a459092254094436aae9dc1f3920f8d871a03053b10377c
 
@@ -84,11 +84,14 @@ RUN conda install -y -q --name neuro numpy=1.14.1 \
                                      seaborn=0.8.1 \
                                      pytables=3.4.2 \
                                      pandas=0.22.0 \
-                                     nipype=1.0.1 \
+                                     nipype=1.0.3 \
                                      patsy \
     && sync && conda clean -tipsy && sync
 
 COPY [".", "/src/fitlins"]
+
+# User-defined instruction
+RUN echo "$VERSION" > /src/fitlins/fitlins/VERSION
 
 USER root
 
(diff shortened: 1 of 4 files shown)


effigies:
@adelavega This doesn't need thorough review; just want to make sure that it doesn't break things for you.

adelavega:
Yep, I know. I just won't get a chance to test that until tomorrow.

effigies:
Cool. No rush.

effigies:
I was able to run your models all the way through. I had `fmriprep` in the output directory, but that shouldn't make a difference at this point. Merging.

adelavega:
ok cool.

PR #43, opened 2018-05-25
FIX: Include run in derivative names

### LINKED ISSUE #42: Run entity not written in contrast filename
In a dataset w/ 4 runs per task, and only one run selected in the model (run: 1), no run entity was recorded in the contrast files output by `BIDSDataSink`. The second level then can't find the files, since it's using `run` as en entity to `.get` files with.

effigies:
Closes #42


adelavega:
LGTM! Tested and it ran.

PR #50, opened 2018-06-15
Loop over preproc files, instead of raw BOLD files

adelavega:
Currently, we loop over BOLD files to get the TR. This creates an unnecessary dependency on the bold files being there (unless you use datalad), which would be nice to remove. Here, I loop over preproc files instead, and look for a matching sidecar of type='bold'.

This is untested.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -167,16 +167,18 @@ class LoadLevel1BIDSModel(SimpleInterface):
                 block.model['HRF_variables'], mode='sparse', force=True):
             info = {}
 
-            bold_files = layout.get(type='bold',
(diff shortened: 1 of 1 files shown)


fitlins/interfaces/bids.py
```diff
@@ -167,16 +167,18 @@ def _run_interface(self, runtime):
                 block.model['HRF_variables'], mode='sparse', force=True):
             info = {}
 
-            bold_files = layout.get(type='bold',
+            space = layout.get_spaces(type='preproc')[0]
```
effigies: I guess the argument is that any space is as good as any other? I think surface spaces (`.gii`) might end up having a weird effect. Can we add `extensions` to this line?

adelavega:
@effigies done!

PR #52, opened 2018-07-03
PIN: Update pybids 0.6.3, grabbit 0.2.1

effigies:
Prep for releasing 0.0.4.


effigies:
@adelavega The new strategy for pulling in metadata based on the preproc files doesn't seem to work. Can you see if I'm breaking something in these changes?

```
180703-14:37:28,33 workflow WARNING:
         [Node] Error on "fitlins_wf.loader" (/scratch/fitlins_wf/loader)
180703-14:37:28,707 workflow ERROR:
         Node loader failed to run on host 148488af3bc9.
180703-14:37:28,710 workflow ERROR:
         Saving crash info to /work/crash-20180703-143728-neuro-loader-874c5725-458f-4c55-a088-fb348492ffab.pklz
  File "/src/fitlins/fitlins/interfaces/bids.py", line 179, in _run_interface
    TR = layout.get_metadata(fname, type='bold')['RepetitionTime']
KeyError: 'RepetitionTime'
```

adelavega:
From a quick look at it, your changes look good. I'll have to take a closer look.

effigies:
///openfmri/ds000030

adelavega:
Hmm, not sure, can you try debugging just using `BIDSLayout` to see if all the files are being indexed? I'll try the same on another dataset later.

effigies:
Yup, working on that.

effigies:
Oh, it's because my fmriprep outputs aren't under the bids root, I bet.

```Python
In [3]: layout = BIDSLayout([('/data/bids/openfmri/ds000030', 'bids'), ('/data/out/ds000030/derivatives/fmriprep', ['bids', 'derivatives'])]) 

In [8]: layout.get_metadata('/data/bids/openfmri/ds000030/sub-10159/func/sub-10159_task-bart_bold.nii.gz', type='bold')['RepetitionTime']
Out[8]: 2

In [9]: layout.get_metadata('/data/out/ds000030/derivatives/fmriprep/sub-10159/func/sub-10159_task-bart_bold_space-MNI152NLin2009cAsym_preproc.nii.gz', type='bold')
Out[9]: {}
```

effigies:
I think we're going to have to revert #50, unless there's something we can do in pybids/grabbit.

adelavega:
Ugh. I think it's because get_metadata searches up the root. This is actually kind of a bigger issue so I think its worth trying to address it in pybids. Some of these helper `get` functions sort of assume that all indexed files are in the same tree. Given the change to `paths`, it should probably search all the paths, not just the first one?

About #50, that's pretty important one for me...

effigies:
I think the (an?) issue is that `BIDSLayout.get_nearest` does not check multiple roots, it just drives to the root directory and stops.

https://github.com/grabbles/grabbit/blob/ebfdb140043931b510140324b8c3cff73e5fa835/grabbit/core.py#L902-L906

tyarkoni:
I think the issue is actually that the filename in your example is not BIDS valid (or if it is, pybids doesn't know what to do with it). The `bold` part is I believe only allowed at the end, before the extension. So what's happening is that the `type` is not being picked up as `'bold'`, so when you ask for `type='bold'`, it doesn't find anything. If you drop the `type` argument, does it work?

effigies:
> I think the issue is actually that the filename in your example is not BIDS valid (or if it is, pybids doesn't know what to do with it).

This is how derivatives work. pybids seems to handle the file fine.

```Python
In [41]: fname = '/data/out/ds000030/derivatives/fmriprep/sub-10159/func/sub-10159_task-bart_bold_space-MNI152NLin2009cAsym_preproc.nii.gz'

In [48]: layout.files[fname].entities
Out[48]: 
{'modality': 'func',
 'space': 'MNI152NLin2009cAsym',
 'subject': '10159',
 'task': 'bart',
 'type': 'preproc'}
```

And no, dropping the `type` argument does not help.

tyarkoni:
I guess this is a general issue we need to consider for derivatives. As the core spec is currently written, 8.3.3 would seem to preclude a functional image ending with anything other than `_bold.nii[.gz]`. If that's the case, the `preproc` would need to occur earlier than that. But it's not obvious to me how that would be handled properly given the `keyword-value` pattern that's separated by underscores...

effigies:
Why does `preproc` need to be earlier? Is that not the type?

tyarkoni:
Okay well if derivatives are allowed to ignore naming conventions stipulated in core BIDS, we're probably going to need to redefine a lot of entities in the BIDS config inside the derivatives config. E.g., `type` will need to relax the stipulation that it comes before a `'.'` in the regex.

effigies:
Okay, so your position is that it *should* be `type=='bold'`, at least as far as `pybids` is concerned?

tyarkoni:
Oh, I see. If you want `preproc` to be the type, that's cool. But then what is the `type='bold'` argument expected to do? Does it work if you do `type='preproc'`?

effigies:
The goal of `type='bold'` is to find the `_bold.json` file that otherwise matches this file. Those exist in the original BIDS directory, while the `_preproc.nii.gz` is in the fmriprep derivatives.

The reasoning for this (#50) was that @adelavega didn't want to require users to have the original BOLD files just for the sake of querying the metadata associated with them.

chrisfilo:
The fact that `preproc` is not a key value pair is ugly. One way of dealing with this is turning it into key/value pair with key such as `srctype` (I'm open for suggestions).

As for pybids support for derivatives it will likely involve some changes. There are two use cases derivatives in /derivatives/<pipeline_name> and derivatives in as standalone datasets. In the first case it will probably be useful to have queries only withing raw data and only within derivatives.

It also might be worth considering functions such as 
`get_srcfile('derivatives/p/sub-01/anat/sub-01_srctype-T1w_preproc.nii.gz')`
 
and 
`get_derivedfiles('sub-01/anat/sub-01_T1w.nii.gz')`

tyarkoni:
Ah, that looks right--sorry I went down a garden path. Should be an easy fix, but I'm busy with a grant till Thursday. Mind taking a crack at it (or @adelavega)? To maintain efficiency, we should probably add a `get_nearest` argument that controls whether or not to look outside the current hierarchy (default to `False`, probably), and then systematically compare the target file with every other file if `True`. We can reuse the constructed `folders` dict after excluding already inspected directories.

tyarkoni:
@chrisfilo I think if derivative files are allowed to ignore hard BIDS conventions, we're going to run into trouble... I don't have strong feelings about the exact convention, but I would prefer that whatever the core spec says must also apply to derivatives (e.g., if `bold` must go at the end, then that should remain true for derivatives as well).

Re: searching through only some paths, there are probably some non-trivial ways that could be achieved right now (would have to think about it), but if there aren't trivial ways, then I agree we should add that ability. (E.g., one easy way would be to allow `Layout.get()` to take a `path` argument that matches each file's path against a regex).

chrisfilo:
Would the proposed solution (changing `_bold_` to `_srctype-bold_`) also ignore the hard BIDS conventions?

tyarkoni:
On my (fairly cursory) reading of the core spec, I think so, yes!

chrisfilo:
Why is that? Could you elaborate? Are you proposing that BIDS derivatives should not add any new suffixes?

effigies:
@chrisfilo @tyarkoni Just to keep this PR thread on topic to the issues needed to get fitlins/pybids/grabbit in sync, could this discussion move to Slack, the Google doc or the mailing list?

chrisfilo:
Fine. 

Yarkoni - do you want to take it outside? We'll see how tough you really are! Just you and me, main square, at noon - ice cream eating competition. You know the rules.

tyarkoni:
I would love for you to take it outside. And while you're outside, I'll just be here inside--eating all the ice cream.

adelavega:
I'll take a crack at it. To confirm:
`get_metadata` should look for a match in other paths outside of the file's root if nothing is found when looking up the tree.
(By the way, I think it's a problem that we need the main BIDS dataset to get the meta-data for a preproc file. Shouldn't this information we self-contained in the fmriprep derivative so that the main dataset is not required?)

tyarkoni:
@effigies already did it, and it's been merged into grabbit

adelavega:
You snooze you lose, huh?

PR #55, opened 2018-07-03
ENH: Allow models without non-HRF variables

### LINKED ISSUE #54: Bug: Crash if no confounds specified
Fitlins assumes that there will be confounds specified which is not guaranteed. If no confounds are present (i.e. variables in 'variables' but not 'HRF_variables'), an error occurs:

```
  File "/work/src/fitlins/fitlins/interfaces/bids.py", line 191, in _run_interface
    names = [col for col in confounds.columns
AttributeError: 'NoneType' object has no attribute 'columns'
```
Seems like an easy enough fix though.

adelavega:
This fixes #54 
Also, I set force=True for `get_design_matrix` mode='dense'. This is because it's plausible yo'd want to control for something (but not convolve with HRF) something that was read in sparse mode. Thus, you'd want to force all "confounds" to dense mode.


adelavega:
Okay, I tested this, and nistats seems to run just fine.

fitlins/interfaces/bids.py
```diff
@@ -3,7 +3,7 @@
 import json
 import shutil
 import nibabel as nb
-
+import pandas as pd
```
effigies: I don't think you use this?

PR #57, opened 2018-07-27
FIX: Limit NaN imputation and use mean non-zero value

### LINKED ISSUE #56: Missing values for confounds result in incorrect number of volumes
Occasionally, confounds have missing values

For example, for `SherlockMerlin` sub-28 `MerlinMovie` task, the first volume for `FrameWiseDisplacement` is `n/a`. 

Regardless of the reason, we should be prepared to deal with such values.

Currently, the `n/a` value will be skipped (by pybids), and result in an incorrect number of volumes in the confounds file passed to nistats. Resulting in this error: 

```
  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nistats/design_matrix.py", line 351, in make_design_matrix
    'time-frames: %d' % (add_regs.shape[0], np.size(frame_times)))
AssertionError: Incorrect specification of additional regressors: length of regressors provided: 1026, number of time-frames: 1027
```

The "proper" BIDS model solution would be the replace n/as with the mean value using  a transformation, but that seems too cumbersome. Probably best if fitlins does this if any n/as are detected in confounds.

effigies:
* [x] Limit imputations to FD and DVARS variants 
* [x] Impute mean non-zero value
* [x] Warn in report on imputation
* [x] Incorporate INCF/pybids#217

Fixes #56.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,7 +81,7 @@ RUN conda install -y -q --name neuro numpy=1.14.1 \
                                      seaborn=0.8.1 \
                                      pytables=3.4.2 \
                                      pandas=0.22.0 \
-                                     nipype=1.0.3 \
+                                     nipype=1.1.1 \
                                      patsy \
     && sync && conda clean -tipsy && sync
 
--- a/fitlins/__about__.py
+++ b/fitlins/__about__.py
@@ -34,19 +34,23 @@ SETUP_REQUIRES = [
 
 REQUIRES = [
     'nibabel>=2.0',
-    'nipype>=1.0.3',
+    'nipype>=1.1.0',
     'seaborn>=0.7.1',
     'numpy>=1.11',
     'nilearn>=0.4.0',
     'pandas>=0.19',
     'tables>=3.2.1',
     'nistats>=0.0.1a',
-    'grabbit>=0.2.1',
-    'pybids>=0.6.3',
+    'grabbit>=0.2.2',
+    'pybids>=0.6.4',
     'jinja2',
 ]
 
 LINKS_REQUIRES = [
+    'git+https://github.com/INCF/pybids.git@'
+    'a86a0e862e93001536bff28e4c2ba35906824d2c#egg=pybids',
+    'git+https://github.com/nistats/nistats.git@'
+    'ce3695e8f34c6f34323766dc96a60a53b69d2729#egg=nistats',
 ]
 
 TESTS_REQUIRES = [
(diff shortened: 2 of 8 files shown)


adelavega:
👌

adelavega:
My PR is merged. Oh by the way, maybe it's not a bad idea to warn users (or throw an exception) if `n/a`s are found in non-imputable variables?

effigies:
Actually, do we even want to by-default impute any but the expected `n/a`s for the first volume? It seems to me that anything else should go through a transform or raise an error.

adelavega:
I agree. I just think a nice exception would be nice to have.

effigies:
> And to complicate matters further, non-steady-state volumes will push down the "first" volume.

Looks like fMRIPrep does not remove non-steady-state volumes before calculating FD/DVARS, so we don't actually need to account for that. (The nistats interface will need to zero them out, but we should already be doing that.)

```
          CSF  WhiteMatter  GlobalSignal   stdDVARS  non-stdDVARS  \
0  608.277697    84.968374    289.321942        NaN           NaN   
1  -36.020893    -1.145165     -8.573609  10.742792    365.722839   

   vx-wisestdDVARS  FramewiseDisplacement  tCompCor00  tCompCor01  tCompCor02  \
0              NaN                    NaN    0.000000    0.000000    0.000000   
1        11.746696               0.495028   -0.013686    0.028413    0.012123   

   ...   Cosine03  Cosine04  Cosine05  NonSteadyStateOutlier00         X  \
0  ...    0.00000  0.000000  0.000000                      1.0 -0.006812   
1  ...    0.10448  0.104445  0.104403                      0.0  0.000000   

         Y        Z      RotX  RotY  RotZ  
0  0.00000  0.00000  0.000000  -0.0   0.0  
1 -0.03263  0.37969  0.001518  -0.0   0.0  

[2 rows x 32 columns]
```

effigies:
Okay. Modulo some prettifying CSS, this is working for me. @adelavega do you want to give it a shot?

adelavega:
can you do a release/push to docker hub? it'll make it easier for me to test

effigies:
So how does it actually get run for you? Do you call `docker` from neuroscout?

adelavega:
I extend your build with a Dockerfile: https://github.com/neuroscout/neuroscout-cli/blob/master/Dockerfile

(just realized I may be re-installing `fitlins` unnecessarily)

effigies:
0.0.5 is on Docker Hub.

adelavega:
Cheers

PR #61, opened 2018-08-01
FIX: Explicitly create working directory

### LINKED ISSUE #59: Bug: default work_dir value (None) causes pathlib error
>   File "/src/fitlins/fitlins/workflows/base.py", line 57, in init_fitlins_wf
>     BIDSDataSink(base_directory=str(Path(base_dir) / 'reportlets' / 'fitlins'),
>   File "/opt/conda/envs/neuro/lib/python3.6/pathlib.py", line 979, in __new__
>     self = cls._from_parts(args, init=False)
>   File "/opt/conda/envs/neuro/lib/python3.6/pathlib.py", line 654, in _from_parts
>     drv, root, parts = self._parse_args(args)
>   File "/opt/conda/envs/neuro/lib/python3.6/pathlib.py", line 638, in _parse_args
>     a = os.fspath(a)
> TypeError: expected str, bytes or os.PathLike object, not NoneType

The default value for `work_dir` in `init_fitlins_wf` is `None` (and in the CLI argparse). But this is fed into `Path`, which causes the error above.

adelavega:
Fixes #59 by explicitly making a tempdir if no work_dir is specified, and making the reportlets directory.

I like this better in any case, because then all the little tempdirs that nipype make are contained in a single one, which makes it easier to debug.


adelavega:
Yeah I think you might be right about `str` now that I've read up on it a bit more.

adelavega:
Otherwise, I tested a model with `FramewiseDisplacement` and it worked

PR #66, opened 2018-08-13
ENH: Set loop_preproc during model loading

### LINKED ISSUE #65: Manipulate `bids.configs` settings
Would you be open to add a CLI flag to set bids.configs settings, such as `loop_preproc=True`?

This would be nice because currently I import fitlins and set bids.configs, prior to calling `run_fitlins`. But we are trying to move towards just making a fully self contained `fitlins` call (either as a Docker container or a pip installed version). This makes it more reproducible, because all inputs and command arguments to fitlins are tracked.

What do you think?

effigies:
This PR does a couple things.

* Set `loop_preproc` config option in `LoadBIDSModel` interface, so that datasets missing original BOLD files can still run. This should make it a FitLins a little cheaper in IO terms, especially if working with DataLad datasets, that now do not need to `get` the originals.
* Relegates bids imports to the insides of interfaces. This is somewhere between aesthetics and encapsulation. It should allow for the `bids.config` changes to be confined to a single interface, but I think that might depend on the nipype plugin whether that's the case.
* Use `bids.BIDSLayout` and `bids.Analysis` rather than `bids.grabbids.BIDSLayout` and `bids.analysis.Analysis`. The former will help avoid the `grabbids`/`layout` organizational issue, and the latter is mostly for symmetry.

Closes #65.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -17,7 +17,6 @@ from nipype.interfaces.base import (
     traits, isdefined
     )
 from nipype.interfaces.io import IOBase
-from bids import grabbids as gb, analysis as ba
 
 from ..utils import dict_intersection, snake_to_camel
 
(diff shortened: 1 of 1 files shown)


fitlins/interfaces/bids.py
```diff
@@ -159,6 +159,8 @@ class LoadBIDSModel(SimpleInterface):
     output_spec = LoadBIDSModelOutputSpec
 
     def _run_interface(self, runtime):
+        import bids
+        bids.config.set_options(loop_preproc=True)
```
effigies: @adelavega Does this look like the right place to do this? My understanding is the time to do it is before `bids.Analysis().setup()` is called. Are there any other cases where it's important to ensure this is set?

adelavega: That seems right to me. The actual function that uses this is `load_variables` so as long as its done prior to that, it should work.

PR #69, opened 2018-08-20
FIX: Compare run entities as int

effigies:
This is a quick hack, pending the resolution of INCF/pybids#236 and INCF/pybids#237.


effigies:
Hack obviated by #70.

PR #71, opened 2018-08-21
ENH: Add ds000117 model

effigies:
Based on the following OpenFMRI model:

#### `model_key.txt`
```
model001 Basic
```

#### `models/model001/condition_key.txt`
```
task001 cond001 scrambled faces
task001 cond002 real faces
```

#### `models/model001/task_contrasts.txt`
```
task001 face_vs_scram -1 1
```


effigies:
@tyarkoni I'm going to test this model once I get the ds117 preprocessed. Want to have a quick look at whether this makes sense?

tyarkoni:
All looks good to me, at least given the spec as it stands right now. The only thing we might have to rework at some point is the handling of the `variables`, pending the discussion in #234 (which hopefully we can make a decision on while we're all here).

PR #83, opened 2018-10-31
[ENH/DOC/BEP002] Collection of changes from a Tour de Fitlins

AdinaWagner:
@yarikoptic:  this is a collection of changes some of which could potentially be adopted as is (we tried to make commits atomic), dropped, or just taken as ideas -- we added some comments which could be useful while RFing to support BEP002 model specification and BIDS Derivatives.  

Commits to probably not bother about at all: (we just wanted to have them committed to get "clean")
- TEMP: disabled code pieces which break and we seemed to not need them 5c66b11cbcfa6ecfff40eb223c7fabc7d4d78fb9
- TEMP+BK: attempt to fixup contrasts handling to be used with current pybids which broke API (again) f42fc030532d798f3f327717e1029e91f890c4fb

Our use case:
- TEMP: hardcode + comment on our case where input dataset is the "derivatives" dataset fe5d929d3189f19d4b57d76ed67d5d67761ce07a

To possibly adopt (as is or with changes):
- ENH+BF: make exception message more informative when no BOLD files are found 2ad43f430bb85b71549296c059a9d33efb458e95
- BEP002: use func.*_bold for analysis (not preproc/) + comments on how to improve 546ec0fb36fad09abf7333be4d0eaed2d4fb2872
- DOC: comment about necessity to get cmdline args into BIDSModel while considering participants 2d47264ba610ae44ff42859348366b6d16f144b0
- ENH: --pdb cmdline option to disable multiproc and fire up pdb session upon failed execution 78925838ce6925733b69b2a500d519df96e622d3 -- it somewhat has limited utility since the majority of execution happens in nipype but it helps to debug early failures

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -115,6 +115,10 @@ def run_fitlins(argv=None):
 
     subject_list = None
     if opts.participant_label is not None:
+        # Note:  this might not work correctly since no include/exclude etc
+        # options are provided so the BIDSModel created and used within that
+        # function would do different treatment than the one created as a
+        # part of the workflow
         subject_list = bids.collect_participants(
             opts.bids_dir, participant_label=opts.participant_label)
 
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -97,6 +97,8 @@ def get_parser():
                          help='maximum number of threads across all processes')
     g_perfm.add_argument('--debug', action='store_true', default=False,
                          help='run debug version of workflow')
+    g_perfm.add_argument('--pdb', action='store_true', default=False,
+                         help='run pdb on failure. Also will use Linear nipype plugin instead of MultiProc')
 
     g_other = parser.add_argument_group('Other options')
     g_other.add_argument('-w', '--work-dir', action='store', type=op.abspath,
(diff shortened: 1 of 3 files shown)


adelavega:
It's not clear what is going wrong with the contrast handling code? pybids 0.7 is very unstable right now, we haven't made much progress implementing what we discussed in the BIDS Model meeting (which will go into 0.7). I'd recommend rolling back to pybids 0.6.5 and the previous fitlins release as that was working for me.

adelavega:
I guess part of the problem is that 0.7 also includes changes related to BIDS-Derivatives (e.g. `suffix=bold` vs `type=preproc`)... so unfortunately teasing the two apart is a challenge.

effigies:
Yeah... the pybids status is a bit of a nightmare, but hopefully we can chew at this from both ends. I've just released an fMRIPrep that should produce Derivatives-RC1-conformant derivatives, and will do a mass run on ds000{030,114,117}. While that's running, I'm going to update the models to conform to the spec.

With those two inputs, hopefully we'll be able to push at both FitLins and pyBIDS and find where all the breaks are happening.

@AdinaWagner @yarikoptic I'll try to look at this, but I'm not sure how easy this will be to review with everything in a state of flux. Sorry I'm not in a position to make contributing easy just yet.

effigies:
@yarikoptic @AdinaWagner If there are any changes from here that didn't get reworked into another PR, please open a new PR with them.

PR #86, opened 2018-12-12
WIP: Dynamic multi-level support

### LINKED ISSUE #64: Handle models with >2 levels
#30 dropped support for 3+ level models. To fix this, it will be easiest to find an example model to support.

adelavega:
Support for dynamically generating higher-level models depending on the number of levels in BIDS StatsModel.

So far, this almost runs so it seems pretty promising. It also cuts down on the complexity of the worksflow since a lot of elements are re-used. 

Closes #64.
Closes #87.

### DIFF
--- a/fitlins/workflows/base.py
+++ b/fitlins/workflows/base.py
@@ -20,10 +20,10 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
     if model is not None:
         specs.inputs.model = model
 
-    all_models = specs.run().outputs.model_spec
-    if not all_models:
+    model_dict = specs.run().outputs.model_spec
+    if not model_dict:
         raise RuntimeError("Unable to find or construct models")
-    if isinstance(all_models, list):
+    if isinstance(model_dict, list):
         raise RuntimeError("Currently unable to run multiple models in parallel - "
                            "please specify model")
 
@@ -34,7 +34,7 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
     loader = pe.Node(
         LoadBIDSModel(bids_dir=bids_dir,
                       derivatives=derivatives,
-                      model=all_models),
+                      model=model_dict),
         name='loader')
 
     if exclude_pattern is not None:
@@ -83,7 +83,7 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
         ])
 
     models = []
-    for ix, step in enumerate(loader.run().outputs.steps):
+    for ix, step in enumerate(step['Level'] for step in model_dict['Steps']):
         # Set up elements common across levels
 
         #
--- a/fitlins/workflows/base.py
+++ b/fitlins/workflows/base.py
@@ -168,6 +168,7 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
 
             wf.connect([
                 (loader, select_entities, [('entities', 'inlist')]),
+                (loader, model, [('session_info', 'session_info')]),
                 (select_entities, getter,  [('out', 'entities')]),
                 (getter, model, [('bold_files', 'bold_file'),
                                  ('mask_files', 'mask_file')]),
@@ -195,7 +196,6 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
 
         wf.connect([
             (loader, select_contrasts, [('contrast_info', 'inlist')]),
-            (loader, model, [('session_info', 'session_info')]),
             (select_contrasts, model,  [('out', 'contrast_info')]),
             (model, collate, [('contrast_maps', 'contrast_maps')]),
             (collate, plot_contrasts, [('contrast_maps', 'data')]),
--- a/fitlins/workflows/base.py
+++ b/fitlins/workflows/base.py
@@ -91,17 +91,25 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
         # various helper nodes to select the correct portions of the model
         #
 
+        level = 'l{:d}'.format(ix + 1)
+
         select_entities = pe.Node(
-            niu.Select(index=ix), name='select_{}_entities'.format(step))
+            niu.Select(index=ix),
+            name='select_{}_entities'.format(level),
+            run_without_submitting=True)
 
         select_contrasts = pe.Node(
-            niu.Select(index=ix), name='select_{}_contrasts'.format(step))
+            niu.Select(index=ix),
+            name='select_{}_contrasts'.format(level),
+            run_without_submitting=True)
 
         # Squash the results of MapNodes that may have generated multiple maps
         # into single lists.
         # Do the same with corresponding metadata - interface will complain if shapes mismatch
-        collate = pe.Node(MergeAll(['contrast_maps', 'contrast_metadata']),
-                          name='collate_{}'.format(step))
+        collate = pe.Node(
+            MergeAll(['contrast_maps', 'contrast_metadata']),
+            name='collate_{}'.format(level),
+            run_without_submitting=True)
 
         #
         # Plotting
@@ -110,7 +118,7 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
         plot_contrasts = pe.MapNode(
             GlassBrainPlot(image_type='png'),
             iterfield='data',
-            name='plot_{}_contrasts'.format(step))
+            name='plot_{}_contrasts'.format(level))
 
         #
         # Derivatives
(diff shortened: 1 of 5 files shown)


adelavega:
Note that I'm running `Analysis` outside of the main workflow (to get # of steps), which seems okay since its a bottleneck anyways.

adelavega:
Update: Seems to work for two level models. Now need to test on higher level models. 
In particular, `l1_metadata` is confusing me, there may be a better way to handle that. Not sure how to propagate contrast metadata to higher levels

effigies:
Yeah, keeping track of the metadata is the worst part of all of this.

adelavega:
Yeah, I suppose nipype is not really natively design with this in mind.

Right now it seems what we're doing is for every list of files (e.g. contrast_maps), passing a parallel meta-data entity list.

An alternative would be to pair everyfile w/ metadata as a tuple, and always be expecting that. Could be kind of kludgy/ugly though.

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
-        (select_l1_entities, getter,  [('out', 'entities')]),
-
-        (loader, select_l1_contrasts, [('contrast_info', 'inlist')]),
 
-        (loader, l1_model, [('session_info', 'session_info')]),
-        (getter, l1_model, [('bold_files', 'bold_file'),
-                            ('mask_files', 'mask_file')]),
-        (select_l1_contrasts, l1_model,  [('out', 'contrast_info')]),
+        select_entities = pe.Node(
+            niu.Select(index=ix), name='select_{}_entities'.format(step))
```
effigies: I believe it's theoretically possible for the same level to occur multiple times in a BIDS model.
```suggestion
            niu.Select(index=ix), name='select_l{:d}_entities'.format(ix + 1))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
 
-        (getter, l1_metadata, [('entities', 'base_dict')]),
-        (l1_model, l1_metadata, [('contrast_metadata', 'dict_list')]),
+        select_contrasts = pe.Node(
+            niu.Select(index=ix), name='select_{}_contrasts'.format(step))
```
effigies: Similarly,
```suggestion
            niu.Select(index=ix), name='select_l{:d}_contrasts'.format(ix + 1))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
-        (l1_metadata, l2_model, [('out', 'stat_metadata')]),
-        (select_l2_contrasts, l2_model, [('out', 'contrast_info')]),
-
-        (l2_model, collate_second_level, [('contrast_maps', 'contrast_maps'),
-                                          ('contrast_metadata', 'contrast_metadata')]),
+        # Squash the results of MapNodes that may have generated multiple maps
+        # into single lists.
+        # Do the same with corresponding metadata - interface will complain if shapes mismatch
+        collate = pe.Node(MergeAll(['contrast_maps', 'contrast_metadata']),
+                          name='collate_{}'.format(step))
```
```suggestion
                          name='collate_l{:d}'.format(ix + 1))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
-
-        #
-        # HTML snippets
-        #
-        (loader, ds_model_warnings, [('warnings', 'in_file')]),
-        (select_l1_entities, ds_model_warnings,  [('out', 'entities')]),
+        plot_contrasts = pe.MapNode(
+            GlassBrainPlot(image_type='png'),
+            iterfield='data',
+            name='plot_{}_contrasts'.format(step))
```
```suggestion
            name='plot_l{:d}_contrasts'.format(ix))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
-        # (plot_l2_contrast_matrix, ds_l2_contrasts, [('figure', 'in_file')]),
-
-        (collate_second_level, ds_l2_contrast_plots, [('contrast_metadata', 'entities')]),
-        (plot_l2_contrasts, ds_l2_contrast_plots, [('figure', 'in_file')]),
-        ])
+        ds_contrast_maps = pe.Node(
+            BIDSDataSink(base_directory=out_dir,
+                         path_patterns=contrast_pattern),
+            run_without_submitting=True,
+            name='ds_{}_contrast_maps'.format(step))
```
```suggestion
            name='ds_l{:d}_contrast_maps'.format(ix))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
+
+        ds_contrast_plots = pe.Node(
+            BIDSDataSink(base_directory=out_dir,
+                         path_patterns=contrast_plot_pattern),
+            run_without_submitting=True,
+            name='ds_{}_contrast_plots'.format(step))
```
```suggestion
            name='ds_l{:d}_contrast_plots'.format(ix))
```

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
+        #  Set up higher levels
+        else:
+            model = pe.MapNode(
+                SecondLevelModel(),
+                iterfield=['contrast_info'],
+                name='{}_model'.format(step))
+
+            wf.connect([
+                (models[-1], model, [('contrast_maps', 'stat_files')]),
+                (l1_metadata, model, [('out', 'stat_metadata')]),
```
effigies: This seems like it can't be right for L3+.

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
+                iterfield='data',
+                name='plot_design')
+
+            # Images
+            ds_design = pe.MapNode(
+                BIDSDataSink(base_directory=out_dir, fixed_entities={'type': 'design'},
+                             path_patterns=image_pattern),
+                iterfield=['entities', 'in_file'],
+                run_without_submitting=True,
+                name='ds_design')
```
effigies: All of these things that apply only to L1 should not happen in the loop.

fitlins/workflows/base.py
```diff
@@ -150,160 +76,134 @@ def join_dict(base_dict, dict_list):
...
-        BIDSDataSink(base_directory=out_dir,
-                     path_patterns=contrast_plot_pattern),
-        run_without_submitting=True,
-        name='ds_l2_contrast_plots')
-
-    #
-    # Connections
-    #
+    models = []
+    for ix, step in enumerate(loader.run().outputs.steps):
```
effigies: Missed this... Either `loader` is in the workflow or it's not. I think maybe it makes more sense to use the `all_models['Steps']` dictionary directly.

fitlins/workflows/base.py
adelavega: correct. this is something i haven't figured out yet. only works for 2 levels as of now.

fitlins/workflows/base.py
adelavega: that makes sense

adelavega:
I agree with all you've said. I originally was going by step level number, but then went w/ step name for verbosity, but sticking w/ level numbers is fine by me.

effigies:
@adelavega I think my branch addresses all of my comments on here, if you want to merge that and work from there.

I'm currently getting the following on a 2-level model:

```
Error: Traceback (most recent call last):

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/utils.py", line 99, in nodelist_runner
    result = node.run(updatehash=updatehash)

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 473, in run

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 557, in _run_interface

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 637, in _run_command

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nipype/interfaces/base/core.py", line 511, in run

  File "/src/fitlins/fitlins/interfaces/nistats.py", line 172, in _run_interface
    model.fit(input, design_matrix=design_matrix)

  File "/opt/conda/envs/neuro/lib/python3.6/site-packages/nistats/second_level_model.py", line 170, in fit
    raise ValueError('A second level model requires a list with at'

ValueError: A second level model requires a list with atleast two first level models or niimgs
```

Having trouble debugging it, as it's in a `MapNode`, which seems to break the ability to plop a debugger into the actual failing location.

fitlins/workflows/base.py
```diff
@@ -149,161 +99,121 @@ def join_dict(base_dict, dict_list):
...
-        (loader, select_l2_contrasts, [('contrast_info', 'inlist')]),
-
-        (l1_model, l2_model, [('contrast_maps', 'stat_files')]),
-        (l1_metadata, l2_model, [('out', 'stat_metadata')]),
-        (select_l2_contrasts, l2_model, [('out', 'contrast_info')]),
-
-        (l2_model, collate_second_level, [('contrast_maps', 'contrast_maps'),
-                                          ('contrast_metadata', 'contrast_metadata')]),
+    models = []
+    for ix, step in enumerate(step['Level'] for step in model_dict['Steps']):
```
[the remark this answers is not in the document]
adelavega: Yes although shouldn't we prioritize fitting the model as specified in the file itself? It seems rare that people would intervene on the model using that CLI option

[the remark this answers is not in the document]
effigies: Agreed. Just making a note.

PR #84, opened 2018-11-12
MNT: Pybids 0.7 compatibility

effigies:
This branch tracks restoration of functionality to FitLins, using derivatives from fMRIPrep 1.2.1 and models updated according to the draft BIDS model spec.

The ds114 model will be updated as needed in bids-standard/BIDS-examples#130.

Needed modifications to fMRIPrep derivatives are being tracked in poldracklab/fmriprep#1384.

PR #88, opened 2018-12-18
FIX: Restore report generation

effigies:
This PR has as its goal restoring report generation, and any stylistic/UX fixes that seem appropriate as I work toward that goal.

Currently reports are based on the presence of contrast matrix plots, which were removed in #84, so no reports are generated. Additional concerns:

* [x] Design matrices produced with columns in alphabetical order - change to `spec['Model']['X']` order.
* [x] Restore design matrix correlation plots

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -232,9 +232,8 @@ class LoadBIDSModel(SimpleInterface):
                 # t/aCompCor
                 # We may want to add criteria to include HPF columns that are not
                 # explicitly listed in the model
-                names = [col for col in dense.columns
-                         if col.startswith('non_steady_state') or
-                         col in step.model['x']]
+                names = step.model['x'].copy()
+                names.extend(col for col in dense.columns if col.startswith('non_steady_state'))
                 dense = dense[names]
 
                 # These confounds are defined pairwise with the current volume
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -232,7 +232,7 @@ class LoadBIDSModel(SimpleInterface):
                 # t/aCompCor
                 # We may want to add criteria to include HPF columns that are not
                 # explicitly listed in the model
-                names = step.model['x'].copy()
+                names = [var for var in step.model['x'] if var in dense.columns]
                 names.extend(col for col in dense.columns if col.startswith('non_steady_state'))
                 dense = dense[names]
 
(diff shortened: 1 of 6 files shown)


fitlins/workflows/base.py
```diff
@@ -196,15 +214,15 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, space, exclude_pattern=None,
                 name='{}_model'.format(level))
 
             wf.connect([
-                (models[-1], model, [('contrast_maps', 'stat_files')]),
-                (l1_metadata, model, [('out', 'stat_metadata')]),
-                (model, collate, [('contrast_metadata', 'contrast_metadata')]),
+                (stage, model, [('contrast_maps', 'stat_files'),
+                                ('contrast_metadata', 'stat_metadata')]),
```
effigies: @adelavega I'm honestly not sure how `l1_metadata` worked here. It was always supposed to be this simple, and being able to drop the `l1_metadata` tag (thanks to bids-standard/pybids#319) allows us to do that.

adelavega: Hmm. yeah that makes way more sense to me... I'll try it out

fitlins/interfaces/bids.py
```diff
@@ -232,9 +232,8 @@ def _load_level1(self, runtime, analysis):
                 # t/aCompCor
                 # We may want to add criteria to include HPF columns that are not
                 # explicitly listed in the model
-                names = [col for col in dense.columns
-                         if col.startswith('non_steady_state') or
-                         col in step.model['x']]
+                names = step.model['x'].copy()
```
effigies: How's this? (I haven't tried it. I assume Pandas' columns object supports `in`.)

```suggestion
                names = [var for var in step.model['x'] if var in dense.columns]
```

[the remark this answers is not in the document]
adelavega: What logical groupings? You mean order that you put into the model yourself? I'm fine with trying to preserve the order in any case, just confusing what you're referring to.

adelavega: Should be OK in theory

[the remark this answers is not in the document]
effigies: Yes, just preserving the model writer's specification. I suspect most people will group related variables together and variables of interest first, so that's all I meant by logical groupings.

effigies:
@adelavega If this works for you, I'll merge. 2+ level reports still aren't ready, as I'll need to figure out another way to reconstruct the contrast matrix for those, but won't get to that before tomorrow.

PR #91, opened 2019-01-19
BF: allow for legit derivatives boolean value in a neuroscout ad-hoc …

AdinaWagner:
…handling of lists.

Previously, my command failed with `TypeError: object of type 'bool' has no len()` when I did not specify any -d command line argument. 
(Comments are @yarikoptics, please be mad at him for them)

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -141,7 +141,10 @@ def run_fitlins(argv=None):
 
     derivatives = True if not opts.derivatives else opts.derivatives
     # Need this when specifying args directly (i.e. neuroscout)
-    if len(derivatives) == 1:
+    # god bless neuroscout, but let's make it work for others!
+    if isinstance(derivatives, list) and len(derivatives) == 1:
+        # WRONG AND EVIL to those who have spaces in their paths... bad bad practice
+        # TODO - fix neuroscout
         derivatives = derivatives[0].split(" ")
 
     pipeline_name = 'fitlins'


effigies:
I'm a little worried about this turning into a series of edge cases, when I'd rather just have a sensible API and Python programs calling internally can do a little munging to make sure they produce valid inputs.

So it looks like this should be a list, possibly empty, of paths. An empty list will be interpreted as checking through `<bids-root>/derivatives`. Any caller that has a string that needs splitting should do that first.

@adelavega @yarikoptic @AdinaWagner Does that work for everyone, or is there a use case where that pattern breaks down?

yarikoptic:
Note that this is just a fix to a now broken code which can't run if derivatives aren't specified. So taking it as is first fixes up an awkward exception.

The Bigger evil is the "annotated" neuroscout ad hoc splitting - that over indeed should be gone.

Indeed, in general should be a list but I think it should be possible to say "no derivatives" to be used (despite all best intentions to keep raw, some bids datasets could and do contain properties enough data). Or on the opposite side, to be provided only derivatives.

yarikoptic:
And BTW @effigies, @AdinaWagner is throwing a party today (Hanover, nh), if you come - we could discuss higher level aspects face to face ;-)

effigies:
@yarikoptic I agree with all of this. So I can go ahead and merge, but I'd like to settle on a good long term approach soon.

Depending on when the party is, it's possible it could happen, but we'd need to leave at least an hour before the snow starts in earnest, so I'm a bit skeptical of the feasibility. And I'd have a toddler...

yarikoptic:
I am no longer scared of toddlers! Forgot that snow is coming, we will try to kick you out in time, or you all will get stuck continuing partying at my place until Monday ;-)

AdinaWagner:
Just so that you also get it from me directly, @effigies, you're very welcome to come and bring your family! We start at 5! Occasion is me leaving this place in 2 weeks time, and I want to thank people here for how much I've learned :-)

effigies:
Cool. Anything we should bring?

effigies:
Tagging @AdinaWagner in case just a response doesn't hit your email.

AdinaWagner:
@effigies, sorry for following this up on on a PR ;-) I'm glad you were able to make it! Thanks for the lovely chocolates, I just now had the chance to try them! Hope you made it home safely (btw, they call this a snowstorm? at best, this was snow...). Unfortunately, Conner forgot his backpack (blue one, contains lunch, toothbrush) here. I'll give it to @yarikoptic on Monday and hope it finds it way home swiftly (I'll clear out anything thats perishable, just in case). Or shall I mail it to you?

effigies:
Made it home safely. 🙂 Yeah, I was hearing a lot that it wasn't as bad as predicted, not just around here. I guess it looked like it had more potential.

I can probably come back up sometime this week, weather permitting, for the lunchbox.

PR #92, opened 2019-01-19
BF: Include desc argument in selectors, remove deprecated 'type' key

AdinaWagner:
I ran into a failure due to the deprecation of the "type" keyword. Instead, keyword 'desc' is used in a non-hardcoded fashion. A separate PR contains an additional desc argparse argument with 'preproc' as the default value.


AdinaWagner:
closing this as commit was merged into #95

PR #93, opened 2019-01-19
BF: Allow for no space in the filename, complain if multiple spaces are found

AdinaWagner:
I ran into an error when my repositories files did not contain the space in their name. I don't think it is necessary to have the space keyword to be compulsory. This PR gives an additional option for no specification of space, and will give a warning if several spaces are found.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -203,8 +203,20 @@ class LoadBIDSModel(SimpleInterface):
                 space = ents.pop('space')
             else:
                 # Picks first match
-                space = analysis.layout.get_spaces(suffix='bold',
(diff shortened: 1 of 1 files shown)


tyarkoni:
My recollection from the BIDS-derivatives meeting is that `space` is mandatory, but doesn't have to be included in the filename (i.e., it can be present in the JSON metadata). Looking at RC1 though, this seems ambiguous. @chrisfilo, any insight?

effigies:
Tal, that's my memory as well. Also the implicit space is orig, I think.

AdinaWagner:
I've merged the changes from this PR into #96 as requested by @effigies, I'll leave it open for the sake of the ongoing discussion, but feel free to close this at any point.

PR #94, opened 2019-01-19
ENH: Add 'desc' argument into argparse, previously hardcoded 'preproc…

AdinaWagner:
…' stays default

Already promised in #92. This allows a command line specification of the 'desc' keyword, and will default to the previously hardcoded 'preproc'.


AdinaWagner:
closing this as commit was merged into #95.

PR #126, opened 2019-03-27
enh: allow cifti support

satra:
- [x] first level support
- [x] second level support

just a place holder to ensure that we get this going - will push at least first level model soon.

@effigies - is there a dataset with cifti output that i can use for testing?


effigies:
I believe I have one that @yarikoptic is helping me host. I'll check. The goal with that one is to pull it down with dataad for CircleCI to run.

effigies:
Nope, no CIFTI. Perhaps we can use one of @mgxd's runs?

mgxd:
@satra you should be able to use voice

effigies:
It looks like @oesteban re-ran with CIFTI recently, so as soon as I can figure out how to update the dataset, we should be able to use ours.

satra:
@effigies - can i assume that fitlins will depend on latest nibabel (the one that will include the cifti enhancements)?

effigies:
@satra Yes. My general approach is libraries should support a wide range of upstream libraries, but user-facing apps can be as bleeding edge as you like.

satra:
@effigies - i still need to add tests, so if there is cifti data that would be great. i'll test this on our own data tomorrow morning.

effigies:
It's in the test dataset. ///labs/poldrack/ds003_fmriprep

satra:
@effigies - thanks - where do i tell it to use the cifti version of files rather than the volumetric files. i think it's here: https://github.com/poldracklab/fitlins/blob/master/fitlins/interfaces/bids.py#L275

but is there a way to override this in the fitlins cli ?

satra:
perhaps via augmenting this: https://github.com/poldracklab/fitlins/blob/master/fitlins/cli/run.py#L87

satra:
but `sub-01_task-rhymejudgment_bold.dtseries.nii` does not have a space keyword. so either this is old at this point or i need to think of other clever ways to ask pybids to return these files

effigies:
You should be able to use `--space ''`. And yeah, we'll need to hack in some way to specify to use CIFTI by CLI. Don't know the ideal way just now

effigies:
@satra Is this RTM?

satra:
@effigies - the only change would be to add something that actually tests the cifti analysis (so the if clauses are executed).

is there a docker container within which i can test this locally? (sorry my brain has slowed down!)

effigies:
`poldracklab/fitlins:master`, but you can just `pip install -r requirements.txt ".[all]"`, as we don't really depend on any external tools.

effigies:
I think we might have a querying problem... I don't know what `layout.get()` invocation we'll need to get CIFTI files. `extension='.dtseries.nii'`? There doesn't appear to be a `space-` entity in [recent fMRIPrep outputs](http://datasets.datalad.org/?dir=/labs/poldrack/ds003_fmriprep/sub-01/func).

satra:
@effigies - i think extension is all we can do for now. it would be a hack that gets removed later.

effigies:
However you want to do it is fine, but one option that won't result in a command line option that we'd have to deprecate would be to make `cifti` or `dtseries` a special value for `--space` that instead sets the selector to extension.

effigies:
@satra Merged master. So we should be down to selecting the space or extension.

satra:
@effigies - this would probably be super easy for you to change, hence a ping.  i see in bids.py a reference to `preproc` around lines 240 and 404. i suspect this is where the conditionals for cifti  (i.e. preproc and dtseries would have to go).

also the current errors are coming from nistats and multiproc on nipype. i'll try to replicate the ds003 test on our cluster to see if i can track these down. also does ds003 have cifti output and updated to reflect the derivatives standard?

effigies:
@satra I spent some insomniatic hours last night working on this. There are a ton of little gotchas that need to be dealt with right now here and in PyBIDS. I've got run-level grayordinates being generated, but the CIFTIs are invalid, so once I get that cleaned up, I'll push to a new branch.

satra:
@effigies - sorry to hear that. please get some sleep.

PR #131, opened 2019-04-18
ENH: Produce all available statistical maps from each analysis unit

### LINKED ISSUE #124: First-level outputs
In #48, we switched from z-scores to effect sizes in the first-level outputs. These are more useful at the second level, but not great for visualizing.

In https://github.com/nistats/nistats/pull/183, we added the ability to easily get all output types (['z_score'](https://github.com/effigies/nistats/blob/f57bea97e7f953088bd66f2c3cc10c60631b011a/nistats/contrasts.py#L224-L244), ['stat'](https://github.com/effigies/nistats/blob/f57bea97e7f953088bd66f2c3cc10c60631b011a/nistats/contrasts.py#L168-L195), ['p_value'](https://github.com/effigies/nistats/blob/f57bea97e7f953088bd66f2c3cc10c60631b011a/nistats/contrasts.py#L197-L222), 'effect_size', 'effect_variance'), so which of these are the most useful? Do people have opinions?

cc @satra @mgxd

### LINKED ISSUE #125: Naming of statistical outputs
Currently we're using the nistats names for the output images produced by model fitting: `['z_score', 'stat', 'p_value', 'effect_size', 'effect_variance']`. Since fitlins is theoretically capable of supporting multiple backends, there's no particular reason we need to adhere to these; we should probably standardize on a set of names that we can reuse everywhere. My take on the names:

* `z_score` and `p_value` are clear and self-explanatory.
* `stat`: From a quick glance, it looks like this is either `t` or `F`, depending on the analysis. I think this is problematic, because you can't tell just from the filename what it represents. I suggest renaming maps to `t_stat` or `F_stat` as appropriate.
* `effect_size`: this is technically accurate, but I dislike it because people tend to use the term "effect size" when talking about standardized effects, so there's a risk of confusion. I suggest `beta`, `par_est`, `parameter_estimate`, or just `effect`.
* `effect_variance`: I'm ambivalent about this one; I think just `variance` would probably be fine. It's possible that there will be other things we want to call `variance`, but I think the surrounding context should always make the semantics relatively clear (e.g., an estimate of between-subject variance at a higher level wouldn't have any subject or run id attached).

Tagging @nicholst, @bthirion, @cmaumet, @jbpoline for potential input.

### LINKED ISSUE #133: Higher-level analyses should use effect maps as inputs, not stat maps
This is related to #48, where we discovered that for 2-level models, the higher-level analysis was using first-level `stat` maps as input, rather than `effect` maps. This was fixed (I think) by changing what the first level outputs. However, this fix no longer works when we have 3-level models, because then the 2nd-level models will output stat maps, and the 3rd level (group) will use stat maps instead of effect maps.

Given that we're moving to outputting all maps by default (#124), we will need (a) a way to indicate which maps to pass on between levels and (b) to default to using `effect`.

effigies:
Fixes #124.
Closes #125.
Closes #133.


effigies:
@tyarkoni @adelavega We should discuss what the appropriate outputs to be displayed at each level, and what are the appropriate ones to pass to the next level.

Currently, we are displaying and passing `effect_maps` at the first level, and then `stat_maps` at all levels higher. I'm guessing this was done for 2-level models, and what we may actually want are `effect_maps` at all but the final level, and then `stat_maps` at the final level.

But a thing to note: nistats doesn't permit using effects and variances to perform second-level models; you just get one map and you run an intercept. I believe @bthirion said he uses z-score maps for this reason, but it's not clear to me that this is equivalent.

bthirion:
No, this is not equivalent. I'm actually working on adding mixed-effects models with known firs-level variance to nistats. I have a branch with it. Will send a PR tomorrow.

adelavega:
#133 relates to this

effigies:
@adelavega This works for me on a 3-level model. Do you want to test before I merge, or just go for it?

adelavega:
Looks good from a quick glance, so go ahead and merge and I'll test with the other upcoming changes.

PR #132, opened 2019-04-19
REL: 0.3.0

effigies:
Now's a good time for small annoying bug fixes. It seems unlikely I'll get anything major in before the end of day, although #131 has a chance.

@adelavega I'd prefer not to break neuroscout, so if you can check master, please do.

@mgxd @satra If you've got anything please let me know.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,25 @@
+0.3.0 (April 19, 2019)
+======================
+
+This release restores reports at the second level and higher, and enables isotropic
+smoothing with the nistats backend. Reporting has also been refactored to reduce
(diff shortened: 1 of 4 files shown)


mgxd:
@effigies is sparse acquisition supported with this?

effigies:
@mgxd Sparse acquisition requires PyBIDS to correctly generated design matrices, and PyBIDS master doesn't support sparse acquisition yet, so no. But this version should be compatible with bids-standard/pybids#376, if you want to verify with the latest master.

mgxd:
sounds good, thanks!

effigies:
#131 isn't going to finish today.

PR #135, opened 2019-04-24
ENH: Enable smoothing at any analysis level

### LINKED ISSUE #134: Smoothing at higher levels
Currently, fitlins only supports smoothing the first-level inputs.

However, @tyarkoni suggested smoothing prior to the group level is sometimes preferable (or is it, after the first level, and before the subject level?).  This is mostly to account for individual anatomical variability. It is also computationally lighter.

Given that nistats supports smoothing at higher levels, this should be easy to implement in fitlins.

adelavega:
Closes #134 

Main issue of debate is the CLI argument. We could split it into two arguments, but I think its fine how it is.


adelavega:
Looks like the fail is unrealted:
```
ERROR: Error installing 'file:///src/fitlins': editable mode is not supported for pyproject.toml-style projects. pip is processing this project as pyproject.toml-style because it has a pyproject.toml file. Since the project has a setup.py and the pyproject.toml has no "build-backend" key for the "build_system" value, you may pass --no-use-pep517 to opt out of pyproject.toml-style processing. See PEP 517 for details on pyproject.toml-style projects.
```

adelavega:
https://github.com/pypa/pip/issues/6433

fitlins/cli/run.py
```diff
@@ -96,10 +96,11 @@ def get_parser():
                         help="use BOLD files with the provided description label")
 
     g_prep = parser.add_argument_group('Options for preprocessing BOLD series')
-    g_prep.add_argument('-s', '--smoothing', action='store', metavar="TYPE:FWHM",
+    g_prep.add_argument('-s', '--smoothing', action='store', metavar="TYPE:FWHM:LEVEL",
```
effigies: I would prefer the levels to be named, e.g. `run`, `session`, etc. So instead of `-s iso:5:1`, it would be `-s run:iso:5`.

It doesn't seem likely that we'll want to smooth multiple times, does it?

Dockerfile
```diff
@@ -100,7 +100,7 @@ RUN bash -c "source activate neuro \
     && sync
 
 RUN bash -c "source activate neuro \
-    &&   pip install --no-cache-dir -e \
+    &&   pip install --no-use-pep517 --no-cache-dir -e \
```
effigies: I'm fixing this in #136.

effigies:
Rebased the PEP517 fix into #137 and merge. Please merge/rebase the latest master.

Dockerfile
```diff
@@ -1,10 +1,11 @@
 # Generated by Neurodocker version 0.5.0
 # Timestamp: 2019-04-29 18:39:28 UTC
 # 
+
 # Thank you for using Neurodocker. If you discover any issues
 # or ways to improve this software, please submit an issue or
 # pull request on our GitHub repository:
-# 
+#
```
effigies: The Dockerfile is auto-generated. Can you just revert it? Or you can run:

```Shell
docker run --rm -i -a stdin -a stdout -a stderr \
    kaczmarj/neurodocker:0.5.0 generate docker - \
    < .neurodocker.json > Dockerfile
```

effigies:
This requires resolution again. Sorry. Also, any objection to switching to `LEVEL:TYPE:FWHM`, where `LEVEL` is a string instead of an int?

adelavega:
What kind of strings? How do you refer to the group level?

effigies:
Step names. So run, session, subject, dataset. And group = dataset.

adelavega:
Of course (duh). That works for me.

adelavega:
Okay, cool. This is ready to go. I'm not able to test it right now because for some reason our workstation is down. Give it a go @effigies? Otherwise, I'll test tomorrow if our workstation/server is back online.

fitlins/workflows/base.py
```diff
@@ -57,17 +57,16 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, analysis_level, space,
         name='getter')
 
     if smoothing:
-        smoothing_params = smoothing.split(':', 1)
-        if smoothing_params[0] != 'iso':
-            raise ValueError(f"Unknown smoothing type {smoothing_params[0]}")
-        smoothing_fwhm = float(smoothing_params[1])
+        smoothing_params = smoothing.split(':', 2)
```
effigies: I would suggest that we make `run` the default, so we interpret `iso:5` as `run:iso:5`. This makes it backwards compatible.

```suggestion
        smoothing_params = smoothing.split(':', 2)
        if len(smoothing_params) < 3:
            smoothing_params.insert(0, 'run')
```

fitlins/workflows/base.py
```diff
@@ -57,17 +57,16 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, analysis_level, space,
...
+        if smoothing_params[1] != 'iso':
+            raise ValueError(f"Unknown smoothing type {smoothing_params[1]}")
+        smoothing_fwhm = float(smoothing_params[2])
+        smoothing_level = int(smoothing_params[0])
```
effigies: Shouldn't be an `int`.

```suggestion
        smoothing_level = smoothing_params[0]
```

effigies:
One last thing to consider is what if we have a two-level model with steps "run" and "dataset", but I might for computational reasons say "I want to smooth after the run level, whether that next level is session, subject or dataset", but I don't want to inspect the model and modify my command line for each specific model.

I see a few options:

1) Quietly fail to smooth if the smoothing level doesn't match some model step.
2) Warn/error if the smoothing level doesn't match some model step.
3) Interpret "session" as applying to the first level session or above (and similarly for other levels).
4) Permit `l1`, `l2`, etc. to explicitly indicate a positional step.

Do you have any preference? From a minimal parsing option, 3 is appealing, but I think 4 + 2 will be the least surprising.

fitlins/workflows/base.py
adelavega: If so, wouldn't it make more sense to make it `TYPE:FWHM:LEVEL` ?

effigies: I see what you're saying, but `LEVEL:TYPE:FWHM` feels like the order of significance to me. Do you disagree?

adelavega:
I had similar concerns when I wanted to add negative indexes (e.g. I want to smooth at the last step (`-1`))

I would say 2 + 4 is a sensible plan, although I think its a bit of a corner case. I suspect most ppl will smooth either at the first or last level.

tyarkoni:
What are valid values for type?

effigies:
> What are valid values for type?

Right now, just `iso` for isotropic. SUSAN smoothing is planned for the future.

tyarkoni:
I would probably order the arguments by likelihood of use... that would be `FWHM:LEVEL:TYPE`. Ideally we could have sane defaults, so if someone just says `-s 4`, isotropic smoothing is only applied at the run level.

tyarkoni:
Is the general philosophy to have fewer structured arguments rather than more named arguments? I don't have a strong opinion on this one way or the other, but I guess if we think the order thing is going to be confusing, I don't see a huge problem having `--fwhm`, `--smoothlevel`, and `--smoothmethod` arguments.

effigies:
The issue there is that as soon as we implement pretty much any other smoothing algorithm, isotropic smoothing will cease to be the most sane default. So `-s 4` will probably need to change what it means based on what the best available smoothing algorithm is.

Maybe that's fine, but just from a reproducibility perspective, I'd want identical arguments to produce as similar as possible algorithmic choices.

effigies:
It's mostly the fact that they're so closely related that it seems weird to break them into different options.

Additionally, if there's any chance people will want to smooth multiple times at different levels, the current approach allows us to do something like `--smooth l1:susan:2 dataset:iso:5` or something. (I'm not sure if that's ever something we'd want, but breaking them apart will make that much worse.)

tyarkoni:
One other thing to keep in mind is that smarter algorithms may need additional parameters... e.g., SUSAN has several. I don't know if these need much twiddling in practice, or if the defaults work well pretty much everywhere, but there's something to be said for using a simple, dumb approach as the default, and letting people explicitly indicate if they want to use something more complex.

I have a hard time imagining a scenario where someone would want to smooth twice, so I don't know if that's even worth supporting right off the bat... it just gives people one more researcher degree of freedom to play with something that would be hard to justify on principled grounds.

adelavega:
Realistically, how many smoothing algorithms will we really support anyways? Probably only SUSAN and Iso, right? 

If so, we could just make them two separate arguments with their own sensible defaults. 

Tal, would you think smoothing prior to the group level (as we are), or after the first level is most sensible as a default?

effigies:
Sure, let's do `FWHM:LEVEL:TYPE`, then. I'd check for `param[0] == 'iso'`, though to interpret it as `<FWHM>:l1:iso` and emit a `FutureWarning` that the options have changed. This parameter was added because people needed it, so there might be a few scripts that depend on that behavior that I'd rather not break.

effigies:
My feeling is that the expectation is generally going to be to smooth before any fitting, but I'm okay with violating the expectation as long as it's a defensible default.

adelavega:
I agree that's probably the expectation as well. So I'm fine with that being the default.

tyarkoni:
Agree with all of the above.

tyarkoni:
Actually, just to clarify: my expectation re: smoothing level would be that it happens as early as possible. Where that becomes ambiguous is when the type of input can vary. I.e., ideally, when fitting run-level models, we'd want to smooth the timeseries inputs, and not the output statistical maps—but I'm assuming that for practical reasons, we're currently doing the latter. OTOH, when fitting group-level models where the earliest inputs are subject-level maps, we'd want to smooth the input volumes, and not the output. Not sure what the solution here is... "Smooth the first set of 3D volumes either found or generated during pipeline execution" captures it semantically, but is maybe a bit counterintuitive.

effigies:
The current default is to smooth the BOLD series, not the statistical maps. This PR adds the possibility of moving the smoothing to immediately prior to any stage.

adelavega:
Yes, and in the results I showed you I was smoothing prior to the group level. From what I understand, smoothing prior to the first level would be primarily for noise, whereas prior to the group level would be mostly anatomical heterogeneity. 

And if I remember correctly, we decided to do the latter as its much faster for the number of models we're running.

In the future, it might be best if we could feed in pre-smoothed run-level inputs to fitlins, especially if using SUSAN which is quite slow.

tyarkoni:
Oh, good. That's what I would expect intuitively for a default—smooth the very first set of inputs.

effigies:
@adelavega Can you also merge master when you work on this next? There are tests now, so might as well make sure you're not breaking them.

adelavega:
Also to clarify, should the default level be "l1" but it also accepts named levels?

effigies:
Yes, I think so.

fitlins/cli/run.py
```diff
@@ -96,10 +96,12 @@ def get_parser():
...
 
     g_prep = parser.add_argument_group('Options for preprocessing BOLD series')
-    g_prep.add_argument('-s', '--smoothing', action='store', metavar="TYPE:FWHM",
+    g_prep.add_argument('-s', '--smoothing', action='store', metavar="FWHM:LEVEL:TYPE",
                         help="Smooth BOLD series with FWHM mm kernel prior to fitting. "
                              "Valid types: iso (isotropic); "
-                             "e.g. `--smothing iso:5` will use an isotropic 5mm FWHM kernel")
+                             "Defaults: LEVEL='subject' TYPE='iso';"
+                             "e.g. `--smoothing 5:dataset:iso will use an isotropic 5mm"
+                             "FWHM on subject-level maps, before evaluating the dataset level.")
```
effigies: I had a couple minor comments, in particular the default should be `l1` and the metavar should use square brackets to indicate optional components. I also think we skip around a little bit and want to restructure the help text, so here's the whole thing rewritten:

```Python
    g_prep.add_argument('-s', '--smoothing', action='store', metavar="FWHM[:LEVEL:[TYPE]]",
                        help="Smooth BOLD series with FWHM mm kernel prior to fitting at LEVEL. "
                             "Optional analysis LEVEL (default: l1) may be specified numerically "
                             "(e.g., `l1`) or by name (`run`, `subject`, `session` or `dataset`). "
                             "Optional smoothing TYPE (default: iso) must be one of: `iso` (isotropic). "
                             "e.g., `--smoothing 5:dataset:iso` will perform a 5mm FWHM isotropic "
                             "smoothing on subject-level maps, before evaluating the dataset level.")
```

Feel free to edit for clarity.

fitlins/workflows/base.py
```diff
@@ -59,17 +60,37 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, analysis_level, space,
...
+        else:
+            if len(smoothing_params) > 1:
+                smoothing_level = smoothing_params[1]
+            else:
+                smoothing_level = 'l1'
+
+            if len(smoothing_params) > 2:
+                if smoothing_params[2] != 'iso':
+                    raise ValueError(
+                        f"Unknown smoothing type {smoothing_params[1]}")
```
effigies: Again have a bunch of little comments, so here's a replacement for L63-L80:

```Python
        smoothing_params = smoothing.split(':', 2)
        # Convert old style and warn; this should turn into an (informative) error around 0.5.0
        if smoothing_params[0] == 'iso':
            smoothing_params = (smoothing_params[1], 'l1', smoothing_params[0])
            warnings.warn(
                "The format for smoothing arguments has changed. Please use "
                f"{':'.join(smoothing_params)} instead of {smoothing}.", FutureWarning)
        # Add defaults to simplify later logic
        if len(smoothing_params) == 1:
            smoothing_params.extend(('l1', 'iso'))
        elif len(smoothing_params) == 2:
            smoothing_params.append('iso')

        smoothing_fwhm, smoothing_level, smoothing_type = smoothing_params
        smoothing_fwhm = float(smoothing_fwhm)
        if smoothing_type not in ('iso'):
            raise ValueError(f"Unknown smoothing type {smoothing_type}")

File: fitlins/workflows/base.py [hunk 8]
```diff
@@ -161,6 +182,9 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, analysis_level, space,
 
         level = 'l{:d}'.format(ix + 1)
 
+        if smoothing and (smoothing_level == step or smoothing_level == level):
```
```suggestion
        if smoothing and smoothing_level in (step, level):
```

adelavega:
Okay, pushed those changes. FYI, might be easier in the future if you do something like:
`git fetch origin pull/135/head:enh/smoothing`
you can then make your changes locally and you should have the power to push to my branch.

effigies:
Cool. Wanted to give you the opportunity to disagree though...

adelavega:
I'm a very agreeable guy...

adelavega:
It would be cool if you could suggest an entire commit though. It's a bit too fussy with the web interface.

effigies:
Sure.

PR #146, opened 2019-05-02
DOC: Build API docs on RTD

effigies:
RTD is down, so leaving this open to remind me to check back in in the morning.


effigies:
Ah, working. https://fitlins.readthedocs.io/en/doc-api_docs/api/modules.html

PR #160, opened 2019-05-23
ENH: Add task-vs-baseline contrast to ds003

effigies:
@rciric @kfinc @oesteban I varied from your model, but I think this is equivalent... Please let me know if this looks reasonable.


examples/models/ds000003/models/model-001_smdl.json
```diff
@@ -39,12 +39,21 @@
...
           "Weights": [1, -1],
           "Type": "t"
+        },
+        {
+          "Name": "task_vs_baseline",
+          "ConditionList": [
+            "trial_type.word",
+            "trial_type.pseudoword"
+          ],
+          "Weights": [0.5, 0.5],
```
oesteban: Shouldn't this be
```suggestion
          "Weights": [1, 1],
```

since the blocks do not overlap at all?

effigies: As a rule, positive values should sum to 1, and negative values should sum to -1. This way we're getting the average response of the voxel to either condition, not the total response to both conditions.

I don't have a citation off the top of my head for this rule.

effigies: Just to think through it, imagine three voxels that respond with 0 noise:

Voxel X responds consistently with amplitude 1 to condition A and amplitude 0 to condition B.
Voxel Y responds consistently with amplitude 0 to condition A and amplitude 1 to condition B.
Voxel Z responds consistently with amplitude 1 to condition A and amplitude 1 to condition B.

We fit a beta for A and B, so voxel X gets betas (1, 0), voxel Y gets betas (0, 1) and voxel Z gets betas (1, 1). If instead we estimate the response to either (A and B), voxel X will get a beta of 0.5 (it responds half the time), voxel Y will get 0.5, and voxel Z will get 1. So the equivalent to fitting a single regressor for the combination is taking the mean, not the sum.

Issue #115, opened 2019-02-22
Enable AFNI estimators


effigies:
Related: #100

effigies:
@yarikoptic @tyarkoni As the folks requesting non-nistats estimators, do either of you have any experience using 3dREMLfit and any other AFNI tools that might be needed to perform analysis?

tyarkoni:
I don't... @shotgunosine, are you (or someone else in the group) able to help with this? It would be great to add AFNI support!

yarikoptic:
Me neither, let's ping @agt24 as well or may be even @afni-rwcox himself?

leej3:
Just wondering what this would entail? Would nipype interfaces be sufficient?

I talked with @afni-gangc briefly. Some of the programs that might be useful for voxelwise stats:

3dREMLfit would need 3dDeconvolve first (individual level)

And for group analysis:
3dMEMA is counterpart of FLAME
3dLME for LME
3dMVM for anova/ancova etc
3dttest++ for GLM

effigies:
If we create the design matrix, do we need 3dDeconvolve?

But yes, nipype interfaces and possibly advice on best practices and pitfalls to be aware of.

afni-rwcox:
If you create the design matrix, then you don’t need 3dDeconvolve. However, the format of design matrix we pass around contains extra information in its header that is useful, such as markers for which columns are “of interest” and which are part of the “baseline” (null hypothesis) model. Columns of interest get text labels to help mark them out in the final dataset. Censoring information can be included. And so on.

The standard option to 3dREMLfit (program for linear time series regression) is “-matrix FILENAME”, where FILENAME contains an XML element that encodes all this auxiliary data, as well as the matrix itself. The alternative is the “-matim FILENAME” option, which just takes an ASCII array of numbers. Here’s the help for that option:

-matim M  = Read a standard .1D file as the matrix.
            * That is, an ASCII files of numbers layed out in a
              rectangular array. The number of rows must equal the
              number of time points in the input dataset. The number
              of columns is the number of regressors.
            * Advanced features, such as censoring, can only be implemented
              by providing a true .xmat.1D file via the '-matrix' option.
            * An advanced intelligence could reverse engineer the XML
              format used by the .xmat.1D files, to fully activate all the
              regression features of this software :-)
           ** N.B.: You can use only 'Col' as a name in GLTs ('-gltsym')
                    with these nonstandard matrix input methods, since
                    the other column names come from the '-matrix' file.

[...]

PR #171, opened 2019-08-30
Add AFNI 3dREMLfit for first-level estimation

leej3:
# Summary

Addresses #115

This is still a work in progress. I've created a FirstLevelModel nipype node that can be swapped in for the default fitlins one. To my knowledge, the estimator, which uses AFNI's 3dREMLfit program, can compute any F-tests and t-tests, with voxelwise REML estimation of the ARMA(1,1) temporal correlations (e.g., https://doi.org/10.1038/s41467-019-09230-w. However, there are some things that need to be considered for it to fully integrate with the fitlins workflow. Also, breaking it into separate nipype nodes and using the 3dREMLfit nipype interface rather than a subprocess call would probably be good.

For now. I have overwritten the `fitlins.interfaces.nistats.FirstLevelModel _run_interface` method and instead of using `nistats.first_level_model.FirstLevelModel.fit` I have written my own `reml_fit` as a method of `fitlins.interfaces.afni.FirstLevelModel`. These can be integrated as desired (I'm happy to help with some guidance) or persist as its own node/sub-workflow that can be swapped into the main fitlins workflow as requested by the user (likely by using the json rather than by the CLI change I’ve implemented here).

 I overwrote `fitlins.interfaces.utils.MergeAll` because it checks for an equal number of contrast metadata elements and maps of each category. This allows me to make my way through a lot of the workflow but it just kicks the can down the road to the `CollateWithMetadata` class. The expectation that all elements have the same number is not met when one considers multirow F-tests (by multirow I am referring to F-tests specified with multiple condition weight vectors as part of the BIDS Stats Models specification json). I'm not sure what you want to do with this...

 
# Other notes on the implementation

 + The main component of translation implemented here is to rewrite the design matrix and metadata from fitlins into a form that is consumable by 3dREMLfit. This form (described [here](https://docs.google.com/document/d/1zpujpZYuleB7HuIFjb2vC4sYXG5M97hJ655ceAj4vE0/edit), along with other useful features of the program) embeds the information in the NIML (XML-like) header of the AFNI design matrix file to guide the model fitting performed. It drops the need to specify many of the options of the 3dREMLfit CLI. I have hacked this together using string interpolation: it works but representing the metadata as XML and writing it to the file may be more elegant. I will look into that.

+ I'm not sure of the desired scope of this estimator. Would you prefer all results computed or a minimal output of the regression fit? As implemented, all statistical maps are computed using AFNI tools (3dREMLfit and then 3dPval). An alternative would be to tweak the data fed into 3dREMLfit (generate the 1d 'Y' object computed by nilearn) so that the output can be unmasked and the results maps can instead be computed using `nistats.contrasts.compute_contrast`. Regarding AFNI's computation of the maps...
  * I couldn't spot whether t-tests should be interpreted as 1- or 2-sided in the stat models specification (in the p-value computation). The AFNI team feels strongly about the fact that it should be 2-sided unless well-justified (and by far most FMRI-related cases are really 2-sided analyses)! I have used 3dPval to compute the p-value statistical maps, with the default being 2-sided. (For a 1-sided interpretation, these can be divided by 2)

  * z-score maps (the z statistic) are computed using a recent option added to 3dPval (https://github.com/afni/afni/commit/7b41df79d00bf51357309084bf3f1dad0f358aec).
  * From the nistats source I think Bonferroni correction (for number of contrasts) is being applied during result map computation. This has not been added but can be within the reml_fit method in python  if desired. Though perhaps this should not be a default and should instead be explicit in the stats model specification?

# Possible additional features
+ adding multi-column stimuli (multiple basis functions). This requires only a small change in the code to implement. I will do that.
+ The minimize memory functionality within _run_interface is not implemented. I’m not sure what’s happening with that so would need help.
+ Similarly, caching of the fit function is not implemented.
+ Voxelwise covariates: at the run level, I believe AFNI would use the term voxelwise regressors. This 4D data orthogonalization could be implemented fairly easily with the 3dREMLfit “-dsort” option if required.
+ An Ordinary Least Squares could be computed in the same program run to compare the impact of the REML-estimated serial correlation pre-whitening. Or indeed the model could be fit quickly for a more basic analysis without ARMA, which is time-consuming (though better as documented in the help). Would this be a desired addition? It would likely provide useful comparisons with other non-REML estimators.
+ Many other tweaks to the computation that might be desirable. Perhaps these could be passed in as kwargs to reml_fit and injected into the command string/passed to the nipype 3d_REMLfit interface. Is this desirable or is the current constrained implementation and choices, or a variant thereof, acceptable?
+ For now, I’ve set the program to use a single thread to allow smooth operation of embarrassingly parallel computation of multiple BIDS runs on the same host. If this is used in a cluster setting that should probably be configurable to match resource specifications.
+ Are there other afni tools that would be especially useful for computing BIDS stats models? I'm happy to help out. Though perhaps waiting for the dust to settle after this addition would be good...
In the stats models spec there is a description of how it attempts to expand model computation beyond the typical summary statistics paradigm. Are there currently implementations of this in fitlins or elsewhere? We (@afni-gangc) know of an approach that incorporates both effect estimate and the associated standard error as input for group analysis (3dMEMA in AFNI, FLAME is FSL, and also something in SPM along the same line). 

# Notes on neuroimaging processing
+ fitlins masks the data before fitting the regression. Typical AFNI processing does not do this. A basic extent mask is computed for the time series (a subset of all EPI voxels that have sufficient coverage over time to be able to meaningfully compute the regression and prevent issues during spatial blurring) but data outside the brain is not masked. As such, artifacts that are present and might be a cause for concern can be observed at the single level analysis output. Masking subsequently is applied before group level analyses/multiple comparisons corrections etc. With this said, if a mask file is supplied to the estimator, the 3dREMLfit calculations are done only on the voxels marked as being in the mask.
+ Also see the above note on t-test sidedness.
+ Censoring is an essential part of this sort of analysis. I had a brief chat with @effigies regarding its implementation in fitlins. Overall it seems like fitlins will be going with additional regressor columns in the design matrix to censor specific timepoints. Typically AFNI just removes the appropriate rows, but 3dREMLfit can process the time series dataset/design matrix either way. @afni-rwcox tested the results of both methods (column augmentation and row removal) and other than a few anomalous voxels (voxels identically zero except at the censored time points), the results were identical when fitting an ARMA model. One consideration for the choice would be that less RAM and CPU are required for the row removal method, because some volumes are excluded from the computation - this becomes significant when the fraction of censored time points goes over (say) 20% of the total.
+ I talked with @effigies about examples of reasonable applications of a multirow F-test. I got some useful ones: 
    * modelling the hrf with multiple basis functions (TENT/CSPLIN functions in AFNI, FIR in SPM/FSL). 
    * Main effects and interactions in an ANOVA design if the user wants to specify this
    * a quick and dirty goodness-of-fit test for the full model specification (e.g., do any of the specified conditions show a meaningful effect, worrying if the answer is "no" in the auditory cortex for an auditory task) 

# Queries about where the appropriate place for certain code-chunks.
+ Since AFNI volumewise statistical codes map onto nifti intent codes precisely, a recoder could be implemented in nibabel for this (as I have done here). Would this be a welcome pull request to nibabel?
+ On a more general note, would you accept pull requests for parsing other volumewise attributes to improve the API to BRIKs in nibabel? One bit of functionality that might be nice is for working with AFNI's "buckets". These are datasets that contain heterogeneous data volumes (e.g., beta coefficients and statistics for components in models). Splitting these into a list of homogenous volumes might be a useful method of the AFNIImage class. And implementing `__iter__` for iterating through each element accessible by slicer[...,idx] might be nice...


tyarkoni:
Holy shit, this fantastic, John! Had no idea you were working on this. Will read through the comments and try to answer any questions I can. Excited to use this! (@adelavega, feel free to experiment—it would be awesome to be able to give NeuroScout users the option to use AFNI for estimation).

leej3:
Excellent. Thanks. I’ll work through the feedback next week.

adelavega:
Very awesome. I guess in theory as far as I'm concerned the API should be plug and play interchangeable, correct? I'll take a closer look at the code next week!

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+        (18, "weibull", "Weibull"),
+        (19, "chi", "Chi"),
+        (20, "inverse gaussian", "Invgauss"),
+        (21, "extreme value 1", "Extval"),
+        (22, "p value", "Pval"),
+        (23, "log p value", "LogPval"),
+        (24, "log10 p value", "Log10Pval"),
+    ),
+    fields=("code", "label", "stat_code"),
+)
```
effigies: Pretty sure this is just `nibabel.nifti1.intent_codes`, restricted to the first 24?

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+    design_vals = design.to_csv(sep=" ", index=False, header=False)
+
+    design_mat = f"""\
+        # <matrix
+        # ni_type = "{design.shape[1]}*double"
+        # ni_dimen = "{design.shape[0]}"
+        # RowTR = "{t_r}"
+        # GoodList = "0..{design.shape[0] - 1}"
+        # NRowFull = "{design.shape[0]}"
+        # CommandLine = "Created by fitlins, no command line used"
```
effigies: Maybe?

```suggestion
        # CommandLine = "{' '.join(sys.argv)}"
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+        if events is not None:
+            events = _check_run_tables(run_imgs, events, "events")
+        if confounds is not None:
+            confounds = _check_run_tables(run_imgs, confounds, "confounds")
+
+        # For each run fit the model and keep only the regression results.
+        nistats_flm.labels_, nistats_flm.results_, self.design_matrices_ = ([], [], [])
+        n_runs = len(run_imgs)
+        t0 = time.time()
+        for run_idx, run_img in enumerate(run_imgs):
```
effigies: Is it important to do these in series, or are they independent? If they're independent, we may be able to rewrite this to run just one, and use a `MapNode` to iterate.

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+        )
+    for statsym in statsyms:
+        val = statsym.replace(")", "").split("(")
+        if val == ["none"]:
+            val.append(tuple())
+            intent_info.append(tuple(val))
+        else:
+            params = [x for x in val[1].split(",")]
+            intent_info.append(
+                (STAT_CODES.label[val[0]], tuple([int(x) for x in params if x]))
```
effigies:
```suggestion
                (nb.nifti1.intent_codes.label[val[0]], tuple(int(x) for x in params if x))
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+        Parameters
+        ----------
+        run_imgs : Niimg-like object or list of Niimg-like objects,
+            See http://nilearn.github.io/manipulating_images/input_output.html#inputing-data-file-names-or-image-objects
+            Data on which the GLM will be fitted. If this is a list,
+            the affine is considered the same for all.
+
+        runtime : nipype runtime object
+
+        nistats_flm : nistats.first_level_model.First_Level_Model
```
effigies: I'm a little unclear on the reasoning here. Is this to pass 3dREMLfit parameters that are consistent with nistats defaults? If so, I would on the whole rather expose them to either the FitLins CLI or the BIDS model spec, so that these have the potential to be selected by the user, and not nistats convention.

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+                if run_idx == 0:
+                    remaining = "go take a coffee, a big one"
+                else:
+                    remaining = (100.0 - percent) / max(0.01, percent) * dt
+                    remaining = "%i seconds remaining" % remaining
+
+                sys.stderr.write(
+                    "Computing run %d out of %d runs (%s)\n"
+                    % (run_idx + 1, n_runs, remaining)
+                )
```
effigies: For messages like this, we can use logging, such as:

```Python
from nipype import logging
logger = logging.getLogger('nipype.interface')

logger.debug(f"Computing run {run_idx + 1} out of {n_runs} runs ({remaining})")

File: fitlins/interfaces/afni.py [hunk 7]
```diff
@@ -0,0 +1,637 @@
...
+
+            # Build the experimental design for the glm
+            run_img = check_niimg(run_img, ensure_ndim=4)
+            if design_matrices is None:
+                n_scans = run_img.get_data().shape[3]
```
effigies:
```suggestion
                n_scans = run_img.shape[3]
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+                    nistats_flm.fir_delays,
+                    confounds_matrix,
+                    confounds_names,
+                    nistats_flm.min_onset,
+                )
+            else:
+                design = design_matrices[run_idx]
+            self.design_matrices_.append(design)
+
+            img_mat = run_img.get_data()
```
effigies: `get_data()` is being deprecated. If you want data scaled according to the scale factors:

```suggestion
            img_mat = run_img.get_fdata()
```

If you want the raw data block, as found on-disk:


```suggestion
            img_mat = run_img.dataobj.get_unscaled()
```

If you want the old `get_data()` behavior where what you get is dependent on whether the scale factors are non-identity:


```suggestion
            img_mat = np.asanyarray(run_img.dataobj)
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+            if nistats_flm.signal_scaling:
+                mean = img_mat.mean(axis=nistats_flm.scaling_axis)
+                if (mean == 0).any():
+                    warn(
```
effigies:
```suggestion
                    logger.warning(
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,637 @@
...
+            f"3dREMLfit and 3dPval computation will be performed in: {tmpdir}\n"
+        )
+        sp.check_output(reml_cmd, shell=True, cwd=tmpdir, stderr=sp.STDOUT)
+        sp.check_output(pval_cmd, shell=True, cwd=tmpdir, stderr=sp.STDOUT)
+        sp.check_output(z_cmd, shell=True, cwd=tmpdir, stderr=sp.STDOUT)
+
+        pval_bucket, pval_labels = load_bucket_by_prefix(tmpdir, pvals)
+        zscore_bucket, z_labels = load_bucket_by_prefix(tmpdir, zscores)
+        extra_bucket, extra_labels = load_bucket_by_prefix(tmpdir, extra_vars)
+        stat_bucket, stat_labels = load_bucket_by_prefix(tmpdir, glt)
```
effigies: L332-L361: Is this stuff not already done in [Remlfit](https://nipype.readthedocs.io/en/latest/interfaces/generated/interfaces.afni/model.html#remlfit)? I assumed all of the parsing work should already be done. If it's not, it would make more sense to fix up that interface than to have a separate implementation here.

I guess we need to add a 3dPval interface. We can build that here or in nipype. (It'll get upstreamed eventually.)

fitlins/cli/run.py
```diff
@@ -116,6 +116,10 @@ def get_parser():
     g_other.add_argument('-w', '--work-dir', action='store', type=op.abspath,
                          help='path where intermediate results should be stored')
 
+    g_other.add_argument("--estimator", action="store",type=str,
+        help="provide the name of the estimator you wish to use",
+        default=None, choices=["afni", None])
```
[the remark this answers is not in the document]
leej3: Yes I don't know what's best here. I will let you pick. The model should be independent of the implementation so it shouldn't be put in the model json correct? And yes estimator for each level should to be considered as one adds such options.

fitlins/interfaces/afni.py
leej3: The intent codes in nifti mirror the afni brik stat codes (they were copied from them). I don't know what's more intuitive to describe it as a stat code mapping or to add the brik labels as aliases to the niftit intent code mapping. I think moving this to nibabel would be good though.

fitlins/interfaces/afni.py
leej3: Yes overall I think this needs to be refactored so let me know what organization you think is best. The approach I took was to mirror the fitlins first level model class, overwriting a method from this class and adding the nistats fit class here so that I could tweak its behavior to instead use 3dREMLfit. This minimized rewriting of pre-existing functionality and allowed me to be as consistent as possible with the pre-existing classes. With that said, the logic is a little muddled. 

Things that might make this better:

Converting this to a functional paradigm so that instead of a class this is implemented as a sub workflow of 4 nodes:
1. Parse the inputs from the fitlins pipeline, reusing the other first level model classes' functionality for this
2. Call the 3dREMLfit node (as a map node)
3. Call the 3dPval node (create this here for now for convenience and upstream later)
4. Parse the outputs as required to integrate the workflow into the main fitlins one.

Does this sound more reasonable? I wasn't sure whether break

fitlins/interfaces/afni.py
leej3: map node sounds good

fitlins/interfaces/afni.py
leej3: noted

fitlins/interfaces/afni.py
leej3: i think I just want the scaled data.

fitlins/interfaces/afni.py
leej3: Sounds good. That does seem like 30 lines of superfluous code. I didn't want to complicate things before a refactor. Would you have just defined and executed the single interface node within the class method? 

I can take a hack at writing the 3dPval interface

fitlins/interfaces/afni.py
leej3: very nice

fitlins/interfaces/afni.py
leej3: Or we could inherit from fitlins.interfaces.nistats.FirstLevelModel that has been changed slightly to allow some customization of the model fit.

fitlins/cli/run.py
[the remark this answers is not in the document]
effigies: Yes, this should definitely be here in the CLI, not in the model.

I'll write up an issue about level-dependent parameterization. I don't think we can/should try to solve it here.

fitlins/interfaces/afni.py
effigies: Since they're in NIFTI, I would probably just stick with that, as it's all official and whatnot. But if you wanted to add them to `nibabel.brikhead`, you're welcome to.

fitlins/interfaces/afni.py
effigies: My thought here was to just use the interface, not wrapped in a `Node()`, which means it would just use the working directory of the `Node` that the whole interface is being run in.

leej3: Great, will do.

fitlins/interfaces/afni.py
leej3: The workflow already uses boldfile as an iterfield in a map node. So the iteration is already done intelligently. the remlfit function (just like the nistats.fit function) has the ability to take a list of images though. I can remove that if you wish in the interest of clarity and brevity.

effigies: Okay. Let me have another read through before we refactor.

effigies:
So I began to make code suggestions, but I think it makes more sense to step back and sketch out an overall API.

Your `fitlins.interfaces.afni.FirstLevelModel` is very nearly identical to `fitlins.interfaces.nistats.FirstLevelModel` up to and including the definition of a `nistats.first_level_model.FirstLevelModel` object. That says to me that what we really want there is to split that stuff out into something like a `fitlins.interfaces.nistats.DesignMatrix` interface that can be used directly.

Let's consider some set of abstract interfaces:

```Python
class DesignMatrixInterface(BaseInterface):
    class input_spec(BaseInterfaceInputSpec):
        session_info = traits.Dict(mandatory=True)
        bold_file = File(exists=True, mandatory=True)
    class output_spec(TraitedSpec):
        design_matrix = File()

class FirstLevelEstimatorInterface(BaseInterface):
    class input_spec(BaseInterfaceInputSpec):
        bold_file = File(exists=True, mandatory=True)
        mask_file = File(exists=True)
        smoothing_fwhm = traits.Float(desc='Full-width half max (FWHM) in mm for smoothing in mask')
        design_matrix = File(exists=True, mandatory=True)
        contrast_info = traits.List(traits.Dict)
    class output_spec(TraitedSpec):
        effect_maps = traits.List(File)
        variance_maps = traits.List(File)
        stat_maps = traits.List(File)
        zscore_maps = traits.List(File)
        pvalue_maps = traits.List(File)
        contrast_metadata = traits.List(traits.Dict)
```

Then we could simply say:

```Python
class nistats.DesignMatrix(DesignMatrixInterface):
        import nibabel as nb
        from nistats import design_matrix as dm
        info = self.inputs.session_info
        img = nb.load(self.inputs.bold_file)
        vols = img.shape[3]

        if info['sparse'] not in (None, 'None'):
            sparse = pd.read_hdf(info['sparse'], key='sparse').rename(
                columns={'condition': 'trial_type',
                         'amplitude': 'modulation'})
            sparse = sparse.dropna(subset=['modulation'])  # Drop NAs
        else:
            sparse = None

        if info['dense'] not in (None, 'None'):
            dense = pd.read_hdf(info['dense'], key='dense')
            column_names = dense.columns.tolist()
            drift_model = None if 'cosine_00' in column_names else 'cosine'
        else:
            dense = None
            column_names = None
            drift_model = 'cosine'

        mat = dm.make_first_level_design_matrix(
            frame_times=np.arange(vols) * info['repetition_time'],
            events=sparse,
            add_regs=dense,
            add_reg_names=column_names,
            drift_model=drift_model,
        )

        mat.to_csv('design.tsv', sep='\t')
        self._results['design_matrix'] = os.path.join(runtime.cwd,
                                                      'design.tsv')
        return runtime


class nistats.FirstLevelEstimator(FirstLevelEstimatorInterface):
        import nibabel as nb
        mat = pd.read_csv(self.inputs.design_matrix, sep='\t')
        img = nb.load(self.inputs.bold_file)
        mask_file = self.inputs.mask_file
        if not isdefined(mask_file):
            mask_file = None
            smoothing_fwhm = None
        flm = level1.FirstLevelModel(
            mask_img=mask_file, smoothing_fwhm=smoothing_fwhm)
        flm.fit(img, design_matrices=mat)

        effect_maps = []
        variance_maps = []
        stat_maps = []
        zscore_maps = []
        pvalue_maps = []
        contrast_metadata = []
        out_ents = self.inputs.contrast_info[0]['entities']
        fname_fmt = os.path.join(runtime.cwd, '{}_{}.nii.gz').format
        for name, weights, contrast_type in prepare_contrasts(
                self.inputs.contrast_info, mat.columns.tolist()):
            maps = flm.compute_contrast(weights, contrast_type, output_type='all')
            contrast_metadata.append(
                {'contrast': name,
                 'stat': contrast_type,
                 **out_ents}
                )

            for map_type, map_list in (('effect_size', effect_maps),
                                       ('effect_variance', variance_maps),
                                       ('z_score', zscore_maps),
                                       ('p_value', pvalue_maps),
                                       ('stat', stat_maps)):
                fname = fname_fmt(name, map_type)
                maps[map_type].to_filename(fname)
                map_list.append(fname)

        self._results['effect_maps'] = effect_maps
        self._results['variance_maps'] = variance_maps
        self._results['stat_maps'] = stat_maps
        self._results['zscore_maps'] = zscore_maps
        self._results['pvalue_maps'] = pvalue_maps
        self._results['contrast_metadata'] = contrast_metadata

        return runtime
```

And then your changes boil down roughly to your `_reml_fit()` function, which could now be `afni.FirstLevelEstimator._run_interface`.

I would recommend against using `nistats.first_level_model.FirstLevelModel` objects to store all of the design parameters. If those are things we need, I would rather make them an explicit part of the input specification, and pass them to `FirstLevelEstimatorInterface`s.

Does that seem like a reasonable approach to you?

leej3:
> Does that seem like a reasonable approach to you?

Yes, it looks good. Shall I make a first pass at it or do you want to do it and I can rebase and fix?

A problem I have noticed is that a variable number of output files are required. For example the number of stats_maps, effect_maps etc depend on the number of contrasts specified. This will get slightly more confusing when multi-row F tests (as defined above) are used because the count for each type of map written to disk will also differ.

I don't know whether the length of traited lists can be defined dynamically (based on parsing the model json). Or alternatively contrast could be an iterfield. Making it an iterfield would get the expected map count right (except for the case of multi-row F tests) but doesn't work with how the nistats prepare_contrasts function works. I'm not under the impression this would be better but thought I should flag it at this point rather than later...

effigies:
I can do a first pass on Monday, but if you're planning on working the weekend, you're welcome to give it a shot.

fitlins/workflows/base.py
```diff
@@ -1,12 +1,22 @@
...
+from nipype.pipeline import engine as pe
+from nipype.interfaces import utility as niu
+# from nipype.interfaces import fsl
+from ..interfaces.bids import (
+    ModelSpecLoader, LoadBIDSModel, BIDSSelect, BIDSDataSink)
+from ..interfaces.nistats import FirstLevelModel, SecondLevelModel
+from ..interfaces.visualizations import (
+    DesignPlot, DesignCorrelationPlot, ContrastMatrixPlot, GlassBrainPlot)
+from ..interfaces.utils import MergeAll, CollateWithMetadata
+from ..interfaces.afni import AFNIMergeAll
```
effigies: These are now being imported inside `init_fitlins_wf`. This works around some annoying new nipype behavior.

leej3: good to know

leej3:
The nistats estimator automatically applies scaling. I have done [the same](https://github.com/afni/fitlins/blob/946e64ecd3496844168967db371eb07bc888c911/fitlins/interfaces/afni.py#L89) albeit with a tweak to accommodate the 2d -> 4d switch in data shape. I don't know whether it would be preferable to specify this a transform in the model json instead.

I have assumed that the Tr (required for the afni design matrix) is in the design matrix index `mat.index[1]`. Is that a reasonable assumption?

leej3:
also I can't figure out why test_packaging fails

fitlins/workflows/base.py
effigies: This was what was causing the packaging test failure.

leej3: Aha. That confused me.

effigies:
I don't know

> I have assumed that the Tr (required for the afni design matrix) is in the design matrix index `mat.index[1]`. Is that a reasonable assumption?

Assuming `mat.index` is in seconds not volumes, I'd think `mat.index[1] - mat.index[0]` would be more reliable, as there's always some kind of chance somebody's figured out how to start at `t != 0`. Do you know that it's in seconds? I haven't looked recently.

But if you have the nistats-built design-matrix, why do you need the TR again?

leej3:
> Do you know that it's in seconds? 

The tr in ds_00003 is 2.0, so I assume the unit is seconds.

> But if you have the nistats-built design-matrix, why do you need the TR again?

I have to add it to the AFNI design matrix. session_info is an input to the design matrix node but not to the l1_model node. Rather than add it to the input spec I just wanted to extract it from that design matrix. I'll use the diff just in case. 

so 
```
mat = pd.read_csv(self.inputs.design_matrix, delimiter="\t", index_col=0)
t_r = mat.index[1] - mat.index[0]

**Comment by effigies** [comment]
Ah, right, you need to translate to a new format. I think that's fine. The only potential issue is that there are experimental designs where there is no constant TR, e.g., clustered volume acquisition. We don't support this well yet, but it's a thing to keep in mind. Does AFNI handle situations like that?

**Comment by effigies** [comment]
@leej3 Do we want to discuss this today? Sorry that it's taken so long to get around to it.

**Comment by leej3** [comment]
sounds good. Discuss what/if any bits of functionality are addressed before
merging. Also, its worth flagging that @tyarkoni mentioned that a refactor
might be ideal before merging this...

**Comment by effigies** [comment]
Rerunning the failed test... Hopefully it was just a hiccup.

**Comment by effigies** [comment]
Can you merge/rebase master?

**Comment by leej3** [comment]
Sure

**Comment by leej3** [comment]
sorry I hadn't spotted your commit. You'll have to push that again.

**Comment by effigies** [comment]
> sorry I hadn't spotted your commit. You'll have to push that again.

Which?

**Comment by leej3** [comment]
I can't see it anymore because I had already refreshed the browser. I believe the hash started with 'a'. And the message had the word 'import' in it.

**Comment by effigies** [comment]
Looks like you force-pushed from adf0e378112a6e17065bcea8262c60fe9bb2946c, which is also a fix I had separately made on `master`, so possibly that's it? Anyway, I think we're probably okay.

**Comment by leej3** [comment]
Cool. There's a couple of patches I worked out with @Shotgunosine that we'll add at some point soon.

**Comment by leej3** [comment]
@effigies, are there any other changes you would like for this?

**Comment by effigies** [comment]
Thanks for this! I will look tomorrow.

**Comment by effigies** [comment]
I will look today.

File: fitlins/interfaces/afni.py [hunk 13]
```diff
@@ -0,0 +1,627 @@
...
+
+        # calc smoothness
+        fwhm = afni.FWHMx()
+        fwhm.inputs.in_file = reml_res.outputs.wherr_file
+        fwhm.inputs.out_file = fname_fmt("model", "residsmoothness").replace('.nii.gz', '.tsv')
+        fwhm_res = fwhm.run()
+        fwhm_dat = pd.read_csv(fwhm_res.outputs.out_file,  delim_whitespace=True, header=None)
+        fwhm_dat.to_csv(fwhm_res.outputs.out_file, index=None, header=False, sep='\t')
+
+        #self._results['residual_smoothness'] = fwhm_res.outputs.out_file
```
effigies, line 156: If we're going to have quoted code, can you do something like:

```Python
# XXX: When XYZ happens, we can uncomment this
# <code>
```

effigies (commented):
Marking my progress. Meeting now, so will get back to it later. Feel free to respond now or wait for me to finish.

fitlins/interfaces/afni.py
leej3: @Shotgunosine, are you happy to remove this?

Shotgunosine: Yep, that's fine to get rid of.

Shotgunosine: I'll get rid of the rest of the commented out code as well.

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,624 @@
...
+            stdev_idx = np.where(stdev_bool)[0]
+            # For multirow ftests there will be more than one index
+            effect_idx = np.where(contrast_bool & effect_bool)[0]
+
+            # Append maps:
+            # for each index into the result objects stored in maps, apply a
+            # modifying function as required and then append it to the
+            # appropriate output list type represented by map_list and write
+            # map_list to disk
+            for map_type, map_list, idx_list, mod_func in (
```
effigies, line 294: mod_func doesn't appear to be used?

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def get_stim_labels(self):
+        # Iterate through all weight specifications to get a list of stimulus
+        # column labels.
+        weights = _flatten([x["weights"] for x in self.inputs.contrast_info])
+        return list(set(_flatten([x.keys() for x in weights])))
+
+def extract_volume(imgs, idx, intent_name, fname):
+    img = imgs.slicer[..., int(idx)]
+    intent_info = get_afni_intent_info_for_subvol(img, idx)
```
effigies, line 333: In theory, `slicer` should also slice the header, so a future-safe way to do this would be to just work from the series.

```suggestion
    intent_info = get_afni_intent_info_for_subvol(imgs, idx)
```

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,624 @@
...
+    intent_info = (*intent_info, intent_name)
+
+    outmap = nb.Nifti1Image(img.get_fdata(), img.affine)
```
effigies, line 336: If we can use the image API, there's the possibility of improving efficiency behind the scenes.

```suggestion
    outmap = nb.Nifti1Image.from_image(img)
```

If BRIK/HEADs don't play nicely (this would be a bug to report upstream), then you could also just do:

```suggestion
    outmap = nb.Nifti1Image(img.dataobj, img.affine)
```

If you're just saving the data, there's no need to coerce it to float64.

fitlins/interfaces/afni.py
```diff
@@ -0,0 +1,624 @@
...
+    outmap = set_intents([outmap], [intent_info])[0]
+    outmap.to_filename(fname)
+
+
+class AFNIMergeAll(MergeAll):
```
effigies, line 341: A docstring explaining why this is happening would be helpful (I'm sure I knew, but now I don't), but this can be a new issue.

fitlins/interfaces/nistats.py
```diff
@@ -141,6 +141,10 @@ def dscalar_from_cifti(img, data, name):
 
 
 class FirstLevelModel(NistatsBaseInterface, FirstLevelEstimatorInterface, SimpleInterface):
+    def __init__(self, errorts=False, *args, **kwargs):
+        super(FirstLevelModel, self).__init__(*args, **kwargs)
+        self.errorts = errorts
```
effigies, line 146:
```suggestion
    def __init__(self, *args, **kwargs):
        # Do not error on errorts being passed, but don't try to use it
        # This provides a common API with AFNI's FirstLevelModel
        kwargs.pop("errorts", None)
        super(FirstLevelModel, self).__init__(*args, **kwargs)
```

fitlins/workflows/base.py
```diff
@@ -1,11 +1,13 @@
 from pathlib import Path
 import warnings
+# from nipype.interfaces import fsl
```
effigies, line 3:
```suggestion
```

fitlins/workflows/base.py
```diff
@@ -1,11 +1,13 @@
...
 
 
 def init_fitlins_wf(database_path, out_dir, analysis_level, space,
                     desc=None, model=None, participants=None,
                     smoothing=None, drop_missing=False,
-                    base_dir=None, name='fitlins_wf'):
+                    base_dir=None, name='fitlins_wf',
+                    estimator=None, errorts=False):
```
effigies, line 10:
```suggestion
                    estimator=None, errorts=False,
                    base_dir=None, name='fitlins_wf'):
```

effigies (approved):
Small fixes here. I think this can go in and I'll make some issues for future improvements.

fitlins/interfaces/nistats.py
leej3: I'm curious about this. It's a nice workaround. Is this a typical pattern or just an idiosyncratic hack until a common API is arrived upon?  Should a logging warning be emitted here?

effigies: I consider it a hack. I don't think a warning is a good idea because it's nothing the user can do anything about, so it would just be a note to self, which this comment also serves as. We could add an `XXX` to make it a little more findable if we consider it worth prioritizing.

fitlins/interfaces/afni.py
leej3: hmmm... that would be an error and mean that the effect variance is currently incorrectly the stdev. I'll check this.

fitlins/interfaces/nistats.py
leej3: Good to know. Thanks. I'm happy with it as is.

fitlins/interfaces/afni.py
leej3: The image series are atypical in that header information expected to be similar across nii volumes  are different (using AFNI buckets to prevent several files being written to disk). Since buckets aren't going to change (sensible reduction in output file count), and the nifti standard isn't going to change, this situation seems like an edge case that will persist. So yes, get_afni_intent_info_for_subvol should run on the full series and slicer could be reasonably expected to slice the header.  I'll change get_afni_intent_info accordingly

leej3:
Ok thanks for taking a look. I'll take a pass through and push remaining suggestions:

+ Document MergeAll subclass
+ check that nibabel image api can be used (or switch to using dataobj)
+ propagate changes resulting from expectation that slicer will slice headers too.
+ fix or remove mod_func for results extraction (depending on whether np.square needs to be applied)

leej3:
@effigies 

> Document MergeAll subclass

So I admit the documentation is needed! I am pretty sure that my [previous comment](https://github.com/poldracklab/fitlins/pull/171#issuecomment-536111191) refers to this issue. In this subclass I overwrite the _list_outputs method to avoid the check of equal list lengths across fields. I don't have a sense of whether this violates the stats models IO spec or if there is an alternative solution that is more elegant. Do you have any suggestions?
 
> check that nibabel image api can be used (or switch to using dataobj)

This looks fine. Nothing blew up.

> propagate changes resulting from expectation that slicer will slice headers too.

This isn't required, the code handles either img or imgs fine. I'll leave get_afni_intent_info as is (and the full series is now passed to get_afni_intent_info_for_subvol as you suggested.

> fix or remove mod_func for results extraction (depending on whether np.square needs to be applied)

@Shotgunosine and I spotted this a while back and then forgot to make the fix. He's working through that now.

effigies:
> > Document MergeAll subclass
> 
> So I admit the documentation is needed! I am pretty sure that my [previous comment](https://github.com/poldracklab/fitlins/pull/171#issuecomment-536111191) refers to this issue. In this subclass I overwrite the _list_outputs method to avoid the check of equal list lengths across fields. I don't have a sense of whether this violates the stats models IO spec or if there is an alternative solution that is more elegant. Do you have any suggestions?

I think let's just note that there's an inconsistent API. It may be that there's something in the nistats vs AFNI implementations, or it may be that the models you're using would require a similar change in nistats.

leej3:
Sounds good.

effigies (approved):
Let's do it. Thanks so much for pushing on this!

tyarkoni:
Seconded—this is awesome!

PR #177, opened 2019-09-20
FIX: Retrieve RepetitionTime from BOLD file metadata

### LINKED ISSUE #175: Getting RepetitionTime meta-data from variable entities is unreliable
See https://github.com/bids-standard/pybids/issues/501 for a longer discussion but here's the summary:

The entities for variables (which include meta-data) should not be relied on to get things such as `RepetitionTime`, as pybids officially limits this to core entities (and its sort of a bug that when a model included "regressors" some entities such as `RepetitionTime` leaked in).

effigies:
@adelavega Would you mind testing this branch? I've never seen the failure you get on my test sets.

If it's easy, we can set up a neuroscout installation with a representative model in the tests to make sure we're not breaking things for your use case.

Closes #175.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -274,9 +274,14 @@ class LoadBIDSModel(SimpleInterface):
                             % (', '.join(spaces), space))
                 ents['space'] = space
 
-            # Metadata is now included in entities
-            TR = ents.pop('RepetitionTime')  # Required field in seconds
-            ents.pop('SkullStripped', None)  # Required by spec, but don't complain if missing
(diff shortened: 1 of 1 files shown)


adelavega:
That's probably a good idea, but my guess is that all your test models include confounds. Anything that is read in as a "regressor" in pybids gets the entities associated with it.

And if you have a model with both normal events + regressors, pybids gives you the union of meta-data in `get_design_matrix`.

adelavega:
To be clear, this behavior is not ideal from pybids, but @tyarkoni said he's going to be refactoring the io code at some point soon, so its probably best to just work around it for now.

PR #183, opened 2019-09-24
ENH: Drop missing model inputs

### LINKED ISSUE #176: Handle variables which only occur in some runs
There are cases when you want to model an event that may only occur in certain runs, but not all.

However, nistats cannot handle events which have all missing values.

It would be useful to detect this situation, and only model the variable in the runs it occurs in.
Then, when doing the subject-level model, feed in only those runs that got modeled.

So in some instances, this may be all the runs, or as few as only a single run. 

This is sort of related to the globbing discussion (https://github.com/bids-standard/pybids/issues/487#issuecomment-532873218), but can also occur when you actually know the names you want to model. So without having to solve this problem at the spec-level, I think we can solve it for this subject of cases at the fitlins implementational level.

adelavega:
Fixes #176 

- Behavior is toggled by CLI flag (`--drop-missing`). By default, it crashes.
- In `FirstLevelModel`, empty columns are detected in design matrix.
- In higher level models, dropping missing inputs is handled by `MergeAll`.

### DIFF
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -46,7 +46,8 @@ class FirstLevelModelInputSpec(BaseInterfaceInputSpec):
     session_info = traits.Dict()
     contrast_info = traits.List(traits.Dict)
     smoothing_fwhm = traits.Float(desc='Full-width half max (FWHM) in mm for smoothing in mask')
-
+    drop_missing = traits.Bool(
+        desc='Drop columns in design matrix with all missing values')
 
 class FirstLevelModelOutputSpec(TraitedSpec):
     effect_maps = traits.List(File)
@@ -67,6 +68,9 @@ class FirstLevelModel(NistatsBaseInterface, SimpleInterface):
         from nistats import design_matrix as dm
         from nistats import first_level_model as level1
         info = self.inputs.session_info
+        drop_missing = self.inputs.drop_missing
+        if not isdefined(drop_missing):
+            drop_missing = False
         img = nb.load(self.inputs.bold_file)
         vols = img.shape[3]
 
@@ -81,8 +85,17 @@ class FirstLevelModel(NistatsBaseInterface, SimpleInterface):
         if info['dense'] not in (None, 'None'):
             dense = pd.read_hdf(info['dense'], key='dense')
 
-            # Remove columns with NaNs
-            dense = dense[dense.columns[dense.isna().all() == False]]
+            missing_columns = dense.isna().all()
+            if drop_missing:
+                # Remove columns with NaNs
+                dense = dense[dense.columns[missing_columns == False]]
+            else:
+                if missing_columns.any():
+                    missing_names = ', '.join(
+                        dense.columns[missing_columns].tolist())
+                    raise RuntimeError(
+                        f'The following columns are empty: {missing_names}.'
+                        'Use --drop-missing to drop before model fitting.')
 
             column_names = dense.columns.tolist()
             drift_model = None if 'cosine_00' in column_names else 'cosine'
@@ -134,7 +147,7 @@ class FirstLevelModel(NistatsBaseInterface, SimpleInterface):
                  **out_ents}
                 )
 
-            # If contrast is not computable (i.e. missing column for run)
+            # If contrast is not computable (i.e. column was dropped)
             if weights is None:
                 for map_list in [effect_maps, variance_maps, zscore_maps,
                                  pvalue_maps, stat_maps]:
@@ -169,7 +182,8 @@ class SecondLevelModelInputSpec(BaseInterfaceInputSpec):
     stat_metadata = traits.List(traits.List(traits.Dict), mandatory=True)
     contrast_info = traits.List(traits.Dict, mandatory=True)
     smoothing_fwhm = traits.Float(desc='Full-width half max (FWHM) in mm for smoothing in mask')
-
+    drop_missing = traits.Bool(
+        desc='Drop missing inputs (i.e. contrasts) from previous level.')
 
 class SecondLevelModelOutputSpec(TraitedSpec):
     effect_maps = traits.List(File)
@@ -201,6 +215,11 @@ class SecondLevelModel(NistatsBaseInterface, SimpleInterface):
         smoothing_fwhm = self.inputs.smoothing_fwhm
         if not isdefined(smoothing_fwhm):
             smoothing_fwhm = None
+
+        drop_missing = self.inputs.drop_missing
+        if not isdefined(drop_missing):
+            drop_missing = False
+
         model = level2.SecondLevelModel(smoothing_fwhm=smoothing_fwhm)
 
         effect_maps = []
@@ -223,11 +242,12 @@ class SecondLevelModel(NistatsBaseInterface, SimpleInterface):
         filtered_variances = []
         names = []
         for m, eff, var in zip(stat_metadata, input_effects, input_variances):
-            if eff is not None: ## Add flag
-                if _match(out_ents, m):
-                    filtered_effects.append(eff)
-                    filtered_variances.append(var)
-                    names.append(m['contrast'])
+            if not drop_missing and eff is not None: ## Add flag
+                raise RuntimeError('Missing input')
+            if _match(out_ents, m):
+                filtered_effects.append(eff)
+                filtered_variances.append(var)
+                names.append(m['contrast'])
 
         for name, weights, contrast_type in prepare_contrasts(self.inputs.contrast_info, names):
             # Need to add F-test support for intercept (more than one column)
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -149,9 +149,7 @@ class FirstLevelModel(NistatsBaseInterface, SimpleInterface):
 
             # If contrast is not computable (i.e. column was dropped)
             if weights is None:
-                for map_list in [effect_maps, variance_maps, zscore_maps,
-                                 pvalue_maps, stat_maps]:
-                    map_list.append(None)
+                continue
             else:
                 maps = flm.compute_contrast(
                     weights, contrast_type, output_type='all')
@@ -242,8 +240,6 @@ class SecondLevelModel(NistatsBaseInterface, SimpleInterface):
         filtered_variances = []
         names = []
         for m, eff, var in zip(stat_metadata, input_effects, input_variances):
-            if not drop_missing and eff is not None: ## Add flag
-                raise RuntimeError('Missing input')
             if _match(out_ents, m):
                 filtered_effects.append(eff)
                 filtered_variances.append(var)
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -40,16 +40,14 @@ def prepare_contrasts(contrasts, all_regressors):
 
 class DesignMatrix(NistatsBaseInterface, DesignMatrixInterface, SimpleInterface):
 
-    def __init__(self, drop_missing=False):
-        super(DesignMatrix, self).__init__()
-        self._drop_missing = drop_missing
-
     def _run_interface(self, runtime):
         import nibabel as nb
         from nistats import design_matrix as dm
         info = self.inputs.session_info
         img = nb.load(self.inputs.bold_file)
         vols = img.shape[3]
+
+        drop_missing = bool(self.inputs.drop_missing)
 
         if info['sparse'] not in (None, 'None'):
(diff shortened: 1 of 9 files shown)


adelavega:
Also, I'm wondering how to handle `None`s with traits, as they will fail currently. Does making it a string work?

fitlins/interfaces/nistats.py
```diff
@@ -21,11 +21,19 @@ def prepare_contrasts(contrasts, all_regressors):
...
-        weights = np.array([
-            [row[col] if col in row else 0 for col in all_regressors]
-            for row in contrast['weights']
-            ])
+        # Are any necessary values missing for contrast estimation?
+        missing = any([[n for n, v in row.items()
+                        if v != 0 and n not in all_regressors]
+                       for row in contrast['weights']])
+        if missing:
+            weights = None
```
effigies: If we continue, then the contrast list is just shorter, and you don't need to deal with `None`s.

```suggestion
            continue
```

Would this work? We'll need to make sure that there isn't some other place where we're generating a list that should match 1-1, but no longer is.

adelavega:
I have to test more, but this seems to have worked for me. 

The main related issue is that if only one image per run remains, a second level moel can't be run, and it will crash.

This is the related issue: https://github.com/neuroscout/neuroscout/issues/641
Seemed like it got more complex, but maybe in the meantime we can have the 2nd level model output the input if it only received one (instead of trying to run a model).

adelavega:
If I'm deciphering correctly, as far as fitlins is concerned, we need to add pass through for random effects models, and implemented fixed effect models as well?

The rest of the issue is more about the spec and how to specific random vs fixed effects.

effigies:
Yes... I need to re-read that, but it's also time to make dinner.

adelavega:
Sounds more appealing, tbh.

adelavega:
Allright, cool, this is ready for review. We can handle the fixed/random stuff in another PR, I think. That may end up being more complicated. Well at least the spec stuff, we could probably bang out the pass-through fairly quickly.

fitlins/interfaces/nistats.py
effigies: Was this not a sufficient place to block the production contrasts involving missing regressors?

adelavega:
I reverted the commit about pass through since it seems we need to implement `FEMA` for that.

adelavega:
@effigies if the tests pass, and you approve, this should be ready. 

i'll tackle adding `FEMA` contrasts in another PR (that might stay open for a while, since it will be blocked by nistats + spec changes)

adelavega:
The main conflict is where to put the `drop_missing` input spec for what I think is now called `DesignMatrix`? I'm guessing this change should go into `DesignMatrixInterface`?

effigies:
I think it needs to go in both DesignMatrix and FirstLevelModel, right? I'll review this now.

fitlins/interfaces/nistats.py
```diff
@@ -37,6 +43,9 @@ def _run_interface(self, runtime):
         import nibabel as nb
         from nistats import design_matrix as dm
         info = self.inputs.session_info
+        drop_missing = self.inputs.drop_missing
```
effigies: `bool(Undefined) == False`.

```suggestion
        drop_missing = bool(self.inputs.drop_missing)
```

fitlins/interfaces/nistats.py
```diff
@@ -37,6 +43,9 @@ def _run_interface(self, runtime):
...
+        if not isdefined(drop_missing):
+            drop_missing = False
```
effigies: Then remove these lines.

fitlins/interfaces/nistats.py
```diff
@@ -50,9 +59,26 @@ def _run_interface(self, runtime):
 
         if info['dense'] not in (None, 'None'):
             dense = pd.read_hdf(info['dense'], key='dense')
+
+            missing_columns = dense.isna().all()
+            if drop_missing:
+                # Remove columns with NaNs
+                dense = dense[dense.columns[missing_columns is False]]
+            else:
```
```suggestion
            elif missing_columns.any():
```

fitlins/interfaces/nistats.py
```diff
@@ -50,9 +59,26 @@ def _run_interface(self, runtime):
...
+                if missing_columns.any():
```
```suggestion
```

And de-indent the following lines.

fitlins/interfaces/nistats.py
```diff
@@ -99,18 +125,20 @@ def _run_interface(self, runtime):
...
         for name, weights, contrast_type in prepare_contrasts(
                 self.inputs.contrast_info, mat.columns.tolist()):
-            maps = flm.compute_contrast(weights, contrast_type, output_type='all')
             contrast_metadata.append(
                 {'contrast': name,
                  'stat': contrast_type,
                  **out_ents}
                 )
+            maps = flm.compute_contrast(
+                weights, contrast_type, output_type='all')
```
effigies: Why?

fitlins/interfaces/utils.py
```diff
@@ -17,16 +20,21 @@ def __init__(self, fields=None):
...
         outputs = self._outputs().get()
-        lengths = None
         for key in self._fields:
             val = getattr(self.inputs, key)
-            _lengths = list(map(len, val))
-            if lengths is None:
-                lengths = _lengths
-            elif _lengths != lengths:
-                raise ValueError("List lengths must be consistent across fields")
+            self._calculate_length(val)
```
effigies:
```suggestion
```

fitlins/workflows/base.py
```diff
@@ -269,6 +271,7 @@ def _deindex(tsv):
             name='ds_{}_contrast_plots'.format(level))
 
         if ix == 0:
+            model.inputs.drop_missing = drop_missing
```
effigies: This should be for `design_matrix`. Maybe just set it in the `DesignMatrix()` call?

adelavega:
I think only in `DesignMatrix` as that's where the errors are thrown. For higher levels they are thrown in `MergeAll`. I'm in the process of running a NS model with these new changes.

adelavega:
I'm getting an error I don't understand:

```
Exception occurred in traits notification handler for object:
bold_file = <undefined>
contrast_info = <undefined>
design_matrix = <undefined>
mask_file = <undefined>
smoothing_fwhm = <undefined>
, trait: trait_added, old value: <undefined>, new value: drop_missing
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/trait_notifiers.py",
line 591, in _dispatch_change_event
    self.dispatch(handler, *args)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/trait_notifiers.py",
line 553, in dispatch
    handler(*args)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes
.py", line 1045, in _set_mapnode_input
    setattr(self._interface.inputs, name, newvalue)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/trait_notifiers.py",
line 400, in __call__
    handle_exception(object, trait_name, old, new)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/trait_notifiers.py",
line 180, in _handle_exception
    raise excp
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/trait_notifiers.py",
line 394, in __call__
    self.handler(*args)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/traits/has_traits.py", line
1168, in _trait_added_changed
    if (trait.type == "delegate") and (name[-6:] != "_items"):
```

effigies:
Are you reusing a working directory?

adelavega:
I don't think so. This is happening on `master` too and I restarted my docker instance and I'm not specifying a work dir so I assume it's should be cleared.

adelavega:
No, I'm not. This doesn't seem to happen on `master` fwiw.

adelavega:
Ah, I see what it is, `drop_missing` was going to the first level model, whereas it should to design matrix now.

adelavega:
I set `drop_missing` as a kwarg for `DesignMatrix` and took it out of the input_spec. Does that make sense to you?

effigies:
> I set `drop_missing` as a kwarg for `DesignMatrix` and took it out of the input_spec. Does that make sense to you?

From the nipype perspective, the contents of the outputs will not purely depend on the contents of the inputs, so you might get a situation where you re-run, adding or removing a `--drop-missing` flag, and the outputs from the previous run get reused, rather than recomputed. I think I would prefer it to be an input.

You can still keep the `DesignMatrix(drop_missing=drop_missing)`, though. That will set the trait in the input spec.

adelavega:
Okay, sure, then what I don't understand is why we don't do the same for `MergeAll`. Shouldn't that also be as an input then?

effigies:
Yeah, strictly speaking, probably. It doesn't change the outputs, but instead changes the error condition, so it's a little less problematic. Also by the time things show up there, they will have had to go through DesignMatrix. So... whichever you want.

fitlins/interfaces/nistats.py
adelavega: I think it just made sense to me to do that right before looping over it. No functional difference.

adelavega:
Okay, your suggestions made sense, and I applied them. Regarding the input spec, I'll leave it as we have it. I'll test this out on NS tomorrow.

fitlins/interfaces/utils.py
```diff
@@ -17,16 +20,20 @@ def __init__(self, fields=None):
...
         for key in self._fields:
             val = getattr(self.inputs, key)
-            _lengths = list(map(len, val))
-            if lengths is None:
-                lengths = _lengths
-            elif _lengths != lengths:
-                raise ValueError("List lengths must be consistent across fields")
+            if self._check_lengths is True:
+                self._calculate_length(val)
             outputs[key] = [elem for sublist in val for elem in sublist]
```
effigies: This may be excessive caution, but just thinking about re-running the same object with changed inputs, we might want to reset this value after using it.

```suggestion
            outputs[key] = [elem for sublist in val for elem in sublist]
        self._lengths = None
```

fitlins/interfaces/nistats.py
```diff
@@ -175,25 +205,39 @@ def _run_interface(self, runtime):
...
-                maps[map_type].to_filename(fname)
-                map_list.append(fname)
+            if len(effects) == 1:
+                print("Passing through results from previous model,"
+                      "as only one is available")
+                effect_maps.append(effects[0])
+                variance_maps.append(_variances[0])
+                zscore_maps.append()
+                pvalue_maps.append()
+                stat_maps.append()
```
effigies: I think these three lines will cause problems?

fitlins/interfaces/utils.py
adelavega: Ok sounds good.

fitlins/interfaces/nistats.py
adelavega: Ah, yes, I thought I reverted that commit.

adelavega:
Hmm, there seems to be a minor problem with the outputting of plots now, let me investigate

adelavega:
I'm getting this kind of error on several of the datasinks:

```
191001-23:41:22,372 nipype.workflow ERROR:
         Node plot_l1_contrasts failed to run on host 75d4c3f13c32.
ERROR:nipype.workflow:Node plot_l1_contrasts failed to run on host 75d4c3f13c32.
191001-23:41:22,373 nipype.workflow ERROR:
         Saving crash info to /work/crash-20191001-234122-root-plot_l1_contrasts-29178660-2dfb-4b6a-a11a-ec7bedf17930.txt
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/plugins/multiproc.py", line 269, in _send_procs_to_workers
    num_subnodes = self.procs[jobid].num_subnodes()
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 1195, in num_subnodes
    self._check_iterfield()
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/nipype/pipeline/engine/nodes.py", line 1220, in _check_iterfield
    "in iterfields.") % iterfield)
ValueError: Input data was not set but it is listed in iterfields.
```

Did the default verbosity change? I don't see anything prior to this error aside from pybids warnings.
Looking at the workdir though, it looks like everything ran.

adelavega:
Okay, tested and working. My linter suggested I change `!=` to `is not`  and that screwed things up.

PR #185, opened 2019-09-26
FIX: Recognize cosine columns without underscores

Shotgunosine:
Fmriprep 1.5.0's cosine columns don't have an underscore in them.


effigies:
Sure, though this sounds like a bug in fMRIPrep. I think to be remotely usable, we're going to have to hack around various issues that crop up before the final release of BIDS derivatives.

PR #187, opened 2019-09-27
ENH: Add expert option to load BIDS layouts from database file

### LINKED ISSUE #186: Add command line option for saved bids layout
I'm running on an HPC with a slow file system, and running one fitlins instance per run on the cluster. It'd be great if each of them didn't have to parse the entire input directory. Can we add a database_file parameter to the CLI to shorten that startup time?

Shotgunosine:
Closes #186


effigies:
You're going to want to add it to these places, as well:

https://github.com/poldracklab/fitlins/blob/1e45226a77abd7b2d4366808f800922537400839/fitlins/utils/bids.py#L75

https://github.com/poldracklab/fitlins/blob/4fea54eb1946abba1fbcb8b2407c7665546131d0/fitlins/interfaces/bids.py#L107

https://github.com/poldracklab/fitlins/blob/4fea54eb1946abba1fbcb8b2407c7665546131d0/fitlins/interfaces/bids.py#L230-L231

https://github.com/poldracklab/fitlins/blob/4fea54eb1946abba1fbcb8b2407c7665546131d0/fitlins/interfaces/bids.py#L421

And this layout should be built with derivatives.

Shotgunosine:
Alright, trying to add the database_file option to those interfaces. Is there a preferred way to specify an optional input that should be an existing file path if provide, but be None otherwise?

effigies:
Or similar.

```Python
if database_file is not None:
    interface.inputs.database_file = database_file
```

Shotgunosine:
Something like this should work for specifying it in the input spec?
```
class ModelSpecLoaderInputSpec(BaseInterfaceInputSpec):
    bids_dir = Directory(exists=True,
                         mandatory=True,
                         desc='BIDS dataset root directory')
    database_file = traits.File(exists=True,
                                desc='Optional path to bids database file.')
    model = traits.Either('default', InputMultiPath(File(exists=True)),
                          desc='Model filename')
    selectors = traits.Dict(desc='Limit models to those with matching inputs')
```

effigies:
Yes.

fitlins/workflows/base.py
```diff
@@ -18,7 +18,10 @@ def init_fitlins_wf(bids_dir, derivatives, out_dir, analysis_level, space,
     wf = pe.Workflow(name=name, base_dir=base_dir)
 
     # Find the appropriate model file(s)
-    specs = ModelSpecLoader(bids_dir=bids_dir)
+    if database_file is not None:
+        specs = ModelSpecLoader(bids_dir=bids_dir, database_file=database_file)
+    else:
+        specs = ModelSpecLoader(bids_dir=bids_dir)
```
```Python
specs = ModelSpecLoader(bids_dir=bids_dir)
if database_file:
    specs.inputs.database_file = database_file
```

effigies:
If this works for you, I'm good to merge it. It would be good to add tests at some point, but most of the lines we're currently missing require iterating over BIDS datasets with different properties, which makes it a finicky proposition.

Shotgunosine:
I've currently only tested that the code still works if you don't pass a database file, I'll try to test functionality when passed a database file sometime today and add it to the CircleCI config. I've got an abstract to write today though, so I may not get to it till later on in the week.

effigies:
@Shotgunosine Heads up that I resolved merge conflicts, so you'll want to `pull` before any further work.

fitlins/cli/run.py
```diff
@@ -94,6 +94,10 @@ def get_parser():
                         help='regex pattern or string to ignore files')
     g_bids.add_argument('--desc-label', action='store', default='preproc',
                         help="use BOLD files with the provided description label")
+    g_bids.add_argument('--database_file', action='store', default=None,
```
effigies:
```suggestion
    g_bids.add_argument('--database-file', action='store', default=None,
```

fitlins/cli/run.py
```diff
@@ -94,6 +94,10 @@ def get_parser():
...
+                        help="Caution, this is an Expert level option subject to change!"
```
effigies:
```suggestion
                        help="Caution, this is an Expert level option subject to change! "
```

fitlins/cli/run.py
```diff
@@ -94,6 +94,10 @@ def get_parser():
...
+                             "Path to SQLite databse containing the index for this BIDS dataset."
```
effigies:
```suggestion
                             "Path to SQLite databse containing the index for this BIDS dataset. "
```

effigies:
Requires #197...

fitlins/cli/run.py
```diff
@@ -138,16 +139,16 @@ def run_fitlins(argv=None):
...
             opts.bids_dir, participant_label=opts.participant_label,
-            database_file=opts.database_file)
+            database_path=opts.database_path)
 
     ncpus = opts.n_cpus
     if ncpus < 1:
         ncpus = cpu_count()
 
     plugin_settings = {
-        'plugin': 'MultiProc',
```
adelavega: This is temporary?

Shotgunosine:
Yeah, just makes debugging easier.

Shotgunosine:
@effigies This is pretty much good to go. I think there's an issue with adding the code coverage file for the test I added. Do you think you could take a look at that?

effigies:
Yup. I'll get to this ASAP.

fitlins/cli/run.py
```diff
@@ -87,22 +88,30 @@ def get_parser():
...
+                        help='registered space of input datasets. '
+                             'Empty value for no explicit space.')
     g_bids.add_argument('--force-index', action='store', default=None,
                         help='regex pattern or string to include files')
     g_bids.add_argument('--ignore', action='store', default=None,
                         help='regex pattern or string to ignore files')
     g_bids.add_argument('--desc-label', action='store', default='preproc',
                         help="use BOLD files with the provided description label")
+    g_bids.add_argument('--database-path', action='store', default=None,
+                        help="Caution, this is an Expert level option subject to change! "
```
adelavega: Not sure if this caution is necessary given how few people use this, and how it's overall an early stage project.

fitlins/cli/run.py
```diff
@@ -170,18 +168,53 @@ def run_fitlins(argv=None):
...
 
     work_dir = mkdtemp() if opts.work_dir is None else opts.work_dir
 
+    # Go ahead and initialize the layout database
+    if opts.database_path is None:
+        database_path = Path(work_dir) / 'dbcache'
+        reset_database = True
+        make_layout = True
+    elif Path(opts.database_path).exists():
+        layout = BIDSLayout.load(opts.database_path)
```
adelavega: It might be better to let pybids handle some of this logic.

That is, is path exists, let pybids decide to load the database from it. 
The problem with the current approach is that if someone passes a valid db path, but then changes the CLIs arguments (so that they would actually want to re-index), no error will be thrown. So their CLI options are being silently ignored in favor for the indexing options in the db_path. 

I rather let pybids crash if the options mismatch. Also simplifies this logic.

fitlins/cli/run.py
```diff
@@ -170,18 +168,53 @@ def run_fitlins(argv=None):
...
     os.makedirs(deriv_dir, exist_ok=True)
-
     bids.write_derivative_description(opts.bids_dir, deriv_dir)
 
     work_dir = mkdtemp() if opts.work_dir is None else opts.work_dir
 
+    # Go ahead and initialize the layout database
+    if opts.database_path is None:
+        database_path = Path(work_dir) / 'dbcache'
+        reset_database = True
```
adelavega: Do we need to manually ask pybids to `reset_database`? It will do it on its own if it doesn't file a valid `sqlite` file inside the `database_path`. 

I suppose this is safer though, as it could be a mismatching `dbcache` from another run. 

An alternative here is to clean up the `dbcache` folder on fitlins teardown. (and put it somewhre more hidden, at least make it `.dbcache`). Or use a tmpdir, although @effigies seemed to think that would be a problem (not guaranted to be on disk).

I think the right solution is to use `tempfile.TemporaryDirectory` but set `prefix` to `work_dir/.dbcache`. That way its cleaned up automatically. Or can also  be cleaned up manually with ``cleanup` function.

fitlins/interfaces/bids.py
```diff
@@ -103,8 +102,9 @@ def _run_interface(self, runtime):
         from bids.analysis import auto_model
         models = self.inputs.model
         if not isinstance(models, list):
+            database_path = self.inputs.database_path
             # model is not yet standardized, so validate=False
-            layout = bids.BIDSLayout(self.inputs.bids_dir, validate=False)
+            layout = bids.BIDSLayout.load(database_path=database_path)
```
adelavega: The use of `.load` here makes sense, as we have already "validated" that the CLI options match the db path.

fitlins/interfaces/bids.py
```diff
@@ -211,24 +213,9 @@ class LoadBIDSModel(SimpleInterface):
     def _run_interface(self, runtime):
         from bids.analysis import Analysis
         from bids.layout import BIDSLayout
-        import re
-
-        force_index = [
```
adelavega: wait, I think this is important, because of this:  "If entry looks like `/<pattern>/`, treat `<pattern>` as a regex"

I think pybids doesn't respect the `/` so we need to strip those and preprocess CLI arguments.

fitlins/cli/run.py
effigies: > I suppose this is safer though, as it could be a mismatching `dbcache` from another run.

Yes, that's the purpose here. If somebody passes a db file, then we assume they know that it's valid. If we're creating it, we don't want to take any chances, and will always clear it.



> I think the right solution is to use `tempfile.TemporaryDirectory` but set `prefix` to `work_dir/.dbcache`. That way its cleaned up automatically.

That would be fine, but I tend to think of anything in the scratch directory as disposable anyway.

fitlins/cli/run.py
effigies: :+1:

fitlins/cli/run.py
```diff
@@ -170,18 +168,53 @@ def run_fitlins(argv=None):
...
+                            validate=False,
+                            force_index=opts.force_index,
+                            database_path=database_path,
+                            reset_database=reset_database)
+
+    subject_list = None
+    if opts.participant_label is not None:
+        subject_list = bids.collect_participants(
+            opts.bids_dir, participant_label=opts.participant_label,
+            database_path=database_path)
```
effigies: Should we just pass the layout, rather than instantiating another one in the function?

fitlins/cli/run.py
```diff
@@ -196,7 +229,7 @@ def run_fitlins(argv=None):
         except Exception:
             retcode = 1
 
-    layout = BIDSLayout(opts.bids_dir, derivatives=derivatives)
+    layout = BIDSLayout.load(database_path)
```
effigies: Now that layout is guaranteed to exist, I think we can just reuse it?

```suggestion
```

fitlins/interfaces/bids.py
effigies: Yeah, I think we need to move this to `cli.py`.

fitlins/interfaces/bids.py
```diff
@@ -142,6 +142,8 @@ class LoadBIDSModelInputSpec(BaseInterfaceInputSpec):
                          desc='BIDS dataset root directory')
     derivatives = traits.Either(traits.Bool, InputMultiPath(Directory(exists=True)),
                                 desc='Derivative folders')
+    database_path = Directory(exists=False,
+                              desc='Optional path to bids database directory.')
```
effigies: Here, you're setting this up as an alternative to bids_dir/derivatives, but in `_run_interface`, you're replacing them entirely. We should be consistent about which one we're doing.

fitlins/interfaces/bids.py
```diff
@@ -395,6 +382,8 @@ class BIDSSelectInputSpec(BaseInterfaceInputSpec):
                          desc='BIDS dataset root directories')
     derivatives = traits.Either(True, InputMultiPath(Directory(exists=True)),
                                 desc='Derivative folders')
+    database_path = Directory(exists=False,
+                              desc='Optional path to bids database path.')
```
effigies: Again. Inputs say optional, runtime says replace.

fitlins/cli/run.py
adelavega: That's true. Maybe I'm overthinking this then, and it might actually be useful to have the index there for debugging. 

I suppose there could be an advantage to not re-indexing if you're running fitlins over and over. Maybe we can over-ride the force-re-index behavior for power users?

fitlins/cli/run.py
adelavega: Looks like it

setup.cfg
```diff
@@ -29,7 +29,7 @@ install_requires =
     pandas>=0.19
     tables>=3.2.1
     nistats>=0.0.1b0
-    pybids @ git+https://github.com/bids-standard/pybids.git@4cdcb0d
+    pybids @ git+https://github.com/bids-standard/pybids.git@8db0fe6634866885cd0e8ed45b05f82fd04e3944
```
effigies:
```suggestion
    pybids @ git+https://github.com/bids-standard/pybids.git@24d2032020fd407e5bcbbbf20943726f48717574
```

effigies:
@Shotgunosine Just a reminder that this one's waiting on you.

Shotgunosine:
Yeah, thanks, have not had time, but will try to get to it soon.

fitlins/cli/run.py
Shotgunosine: I deffo want it for debugging, and this way, if they want to reuse a layout saved in a working directory, they can pass the path for it explicitly, otherwise it gets reset in order to avoid nasty surprises.

fitlins/interfaces/bids.py
Shotgunosine: Would you prefer it to be optional or to have `LoadBIDSModel` and `BIDSSelect` only work if passed a `database_path`?

adelavega: Optional at the CLI level for sure. But we could set up a default internal path (don't we already do that?).

fitlins/interfaces/bids.py
Shotgunosine: Agreed

Shotgunosine:
Alright, changes made. Let me know if there's anything else to tweak.

effigies:
Thanks for pushing on this. Today was slow (sick kid), but I'll try to get to this on Monday.

effigies:
LGTM. Thanks for your patience.

PR #188, opened 2019-09-30
RF: Abstract interfaces to simplify swappability

effigies:
This is a first pass at abstracting interfaces, intended to ease #171, which reuses the nistats design matrix for the time being.

The goal is to try to move our interfaces away from the specifics of nistats to the inputs and outputs inherent to the processes at each stage. We may want to further refine the scope, or increase the range of abstract processes. For example, we could reduce the estimator outputs to effects, variances and metadata, and move stat, p and Z maps to a separate interface, which AFNI could implement with `3dPval`.

### DIFF
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -2,10 +2,10 @@ import os
 import numpy as np
 import pandas as pd
 
-from nipype.interfaces.base import (
-    LibraryBaseInterface, SimpleInterface, BaseInterfaceInputSpec, TraitedSpec,
-    File, traits, isdefined
-    )
+from nipype.interfaces.base import LibraryBaseInterface, SimpleInterface, isdefined
+
+from .abstract import (
+    DesignMatrixInterface, FirstLevelEstimatorInterface, SecondLevelEstimatorInterface)
 
 
 class NistatsBaseInterface(LibraryBaseInterface):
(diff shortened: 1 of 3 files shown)


effigies:
From separate discussion with @Shotgunosine, we should consider how to deal with estimators that give you certain features others might not.

For example, nistats provides smoothing in first level model, which drastically reduces the disk consumption (while increasing memory). And I believe remlfit calculates residuals in passing.

For this, we could have pass-through inputs/outputs. So for instance we could have a null smoothing interface for nistats that just passes the original bold file and a smoothing parameter to the estimator. If remlfit doesn't, we would make the smoothing interface a 3dblur (or whatever it's called). Similarly with residuals, nistats wouldn't return any and the next stage would calculate them, while remlfit's next stage would be null.

effigies:
cc @leej3

Shotgunosine:
@leej3 do you understand who the reml pull request would need to be changed to work with these suggestions?

fitlins/interfaces/nistats.py
```diff
@@ -90,6 +69,15 @@ def _run_interface(self, runtime):
         mat.to_csv('design.tsv', sep='\t')
         self._results['design_matrix'] = os.path.join(runtime.cwd,
                                                       'design.tsv')
+        return runtime
+
+
+class FirstLevelModel(NistatsBaseInterface, FirstLevelEstimatorInterface, SimpleInterface):
```
adelavega: should the name of this class remain `FirstLevelModel` (and be identifiable by the fact that it lives in nistats.py) or have a more specific name?

adelavega: I see that the `NistatsBaseInterface` also has a `_pkg` variable (does that have some special NiPype meaning?)

effigies: Yeah, at least for now I'm going to do that. Also other tools aren't obliged to use `FirstLevelModel`, which corresponds to the nistats object.

effigies: > I see that the `NistatsBaseInterface` also has a `_pkg` variable (does that have some special NiPype meaning?)

Yes, that exists to allow run-time detection of whether a Python module is installed and its version. So if I want to modify my behavior based on changes in the underlying library, that's there.

leej3:
Looks good. Should help for reducing redundant code in the remlfit pull request. I'll update tomorrow.

> For example, we could reduce the estimator outputs to effects, variances and metadata, and move stat, p and Z maps to a separate interface, which AFNI could implement with `3dPval`.

3dREMLfit outputs the stat map along with the effect/variance maps. So moving  stat/pval/zscore maps to a separate interface would be fine but I would probably just use that interface rather than rewrite one.

effigies:
@adelavega Any objections to me merging this? I haven't looked at #183 enough to know whether I'd be breaking things there, but I want to get this in so I'm not holding up #171.

I can also review yours and merge it and then deal with any merge conflicts here, but that'll probably be >2 hours.

adelavega:
That's fine, np.

PR #190, opened 2019-10-02
FIX: Set smoothing_fwhm after creating next level. Fixes #189

### LINKED ISSUE #189: Smoothing is being applied at previous level
Looks like `smoothing_level` is off by one, and the `smoothing_fwhm` is actually being passed the level one before we meant to.

adelavega:
Fixes #189 

The `smoothing_fwhm` was applied to model before the next level's model was created. This resulted in the `Subject` level inputs being smoothing rather than `Dataset`


effigies:
Feel free to merge when tests pass.

adelavega:
Cool. Unfortunately looking at the git blame here, I can't find any scapegoats for this one.

Issue #151, opened 2019-05-14
handling multiple runs via fixed effects modeling


satra:
currently fitlins doesn't handle multiple runs from a single participant for a given task/set of contrasts in the traditional neuroimaging way (i.e. using fixed effects).

nistats supports this by passing multiple runs and design matrices to the first level model fit and then using a fixed effects model in estimating each contrast across runs.

example: https://nistats.github.io/auto_examples/02_first_level_models/plot_spm_multimodal_faces.html#sphx-glr-auto-examples-02-first-level-models-plot-spm-multimodal-faces-py

code piece: https://github.com/nistats/nistats/blob/master/nistats/first_level_model.py#L568

there are cases where projects would want to look at run level contrasts and cases where this would be combined. this seems like this needs to be handled at the level of bids model as well as fitlins.

one way to extend the interface is to output/store labels and results in addition to the contrasts for each run. then we could call the hidden function (not a clean approach). 

cc: @tyarkoni @effigies

satra:
@tyarkoni - any thoughts on this re: the model spec? 

i think we can automatically assume that run level does both individual runs and groups runs together (via fixedfx). but if we wanted something to only pull out a contrast from a specific run for group stats, we would need to add run index info somewhere.

i'd like to finish the cifti processing and this together. should i just make the assumption and hack fitlins to support it?

effigies:
If you wanted to do a fixed effects within subject, wouldn't you just make subject the first level?

Not that FitLins can handle concatenating runs as yet, but as far as PyBIDS is concerned, I think that should produce the correct design matrix.

satra:
@effigies - yes, i just don't remember the distinction between run and subject in the model specification. do we have run->subject-> session-> dataset ?

effigies:
So these levels are the units that you're parallelizing across. By having a run level, you're saying estimate each run separately. If you're setting your first level as the subject, then you are aggregating all runs within a subject for the first level.

Right now, you have to go from lower-level to higher, so I don't know that you can do subject -> session. Though it seems obvious you should be able to.

satra:
i have to check where in fitlins you have the separation between run and subject level.  if you have direct pointers, that would help.

as a note, second level in nistats is not subject level, it's group/dataset level.

tyarkoni:
[...]

PR #191, opened 2019-10-02
Add fixed effects - FEMA contrasts

### LINKED ISSUE #170: Handle subject level when there is only 1 run
Related to https://github.com/neuroscout/neuroscout/issues/641

Currently, if you only have one run per subject, you cannot specify a subject level, because it requires at least 2 inputs. 

However, this causes a problem that its difficult to programmatically find subject level maps for analyses with only 1 run per subject (because they will have a `run` entity in the file name, and subject level maps do not).

@tyarkoni suggested in this scenario, fitlins at least copies over the outputs from the run level model, as the output of a "subject" level model.

adelavega:
This PR depends on: https://github.com/nistats/nistats/pull/386

- This adds fixed effects contrasts for "FEMA" contrasts in the BIDS StatsModel. 
- Pass-through for cases where there is only a single effect/variance per subject. This works automatically since nistats fixed effects can accept one input. Fixes #170 
- nistats fixed effects does not output p or z values, because that work required the degrees of freedom from the previous level. we might want to add this in the future, but for now, to deal with this, made these two optional throughout the workflow (e.g. in collate functions)


adelavega:
@effigies no rush on reviewing this, but unless nistats api changes, this looks good to me.

the main point of contention is how to handle the fact that `z` and `p` stats are not output by fixed effects.

I also hard coded fixed effects at the subject level. maybe you will disagree with this? or it can be specified in the CLI? In any case this should be temporary until the StatsModel catches up.

adelavega:
The other point of contention is that nistats fixed effects won't handle smoothing. we can do it ourselves using the masker object (and pass one in), or give a warning. currently, only silently ignoring.

fitlins/interfaces/nistats.py
```diff
@@ -202,36 +204,59 @@ def _run_interface(self, runtime):
...
             contrast_metadata.append(
                 {'contrast': name,
                  'stat': contrast_type,
                  **out_ents})
 
-            design_matrix = pd.DataFrame(
-                {'intercept': weights[weights != 0]})
+            # For now hard-coding to do FEMA at the subject level
+            # Pass-through happens automatically as it can handle 1 input
+            if self.inputs.level == 'Subject':
```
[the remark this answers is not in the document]
adelavega: This is coming from `model_dict` so no, its uppercase. I'm getting this value from the for loop used in workflow creation.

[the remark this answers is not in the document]
adelavega: I original had it lowercase and it didnt work.

adelavega:
Yes, I think this could also be done at the `session` level...

I would also like to put it in the spec for future reproducibility, but @tyarkoni indicated this might not happen for a while, so he suggested hard-coding it for now.

About smoothing, we could very easily do what `nistats` does internally. They use the `NiftiMasker` object. We would simply have to initialize one (with the smoothing parameters), and then transform the images. The reason that Bertrand doesn't make to support it in fixed effects, is he said it would not make sense to do at the subject level so he would rather discourage it heavily. We could follow a similar principle, or leave it up to the user. 
See: https://github.com/nistats/nistats/pull/386#discussion_r330751631

adelavega:
The ds114 test-retest is not part of the CircleCI tests, right? Should it be?

effigies:
> @tyarkoni indicated this might not happen for a while, so he suggested hard-coding it for now.

Can we not just do it? It's a draft spec. If it changes in the future, we adapt.

effigies:
And yeah, we can add ds114, if we upload a preprocessed copy of the data somewhere. I'd prefer with datalad, but whatever.

adelavega:
I can host ds114 on our tacc servers

adelavega:
Allright, I've updated this branch to use: https://github.com/bids-standard/pybids/pull/520

It now handles reading the "FEMA" contrast type from the bids model.
I tested it on two models in two datasets and its seemed to work fine.
Passthrough at the Subject level using "FEMA" with only a single run per subject also worked.
 
(We also have a PR on Neuroscout to make these models)

So all we really need now is to hammer out if my proposed changes to the spec draft are OK.
No rush to review this @effigies, but its ready.

adelavega:
This should also close #151

effigies:
Cancelling the job because it's broken.

adelavega:
@effigies if tests pass, this is ready for a final review. 

To explain a bit more, previously we were computing  a separate l2 model for each contrast, but this is not necessary.

Instead you can make a dummy coded design matrix that represents the identity of each `effect_map`.
| effect_file              | cond1 | cond2 |
|--------------------------|-------|-------|
| cond1_effect_size.nii.gz | 1     | 0     |
| cond2_effect_size.nii.gz | 0     | 1     |
| cond1_effect_size.nii.gz | 1     | 0     |
| cond2_effect_size.nii.gz | 0     | 1     |

You then just translate the weights (represented as dicts), to match the columns:

    {name='cond1', weights=[1, 0], type='t'}

This works for F-tests too:
    {name='omnibus', weights=[[1, 0], [0, 1]], type='F'}

adelavega:
Realized that the same logic was not being applied for `FEMA` contrasts. For those, it's typically a `Dummy` contrast, but in theory it could be more than that, something like: `weights: [1, 1, 0, 0]`. 

That is, you could do a meta-analysis across multiple effects. Maybe its rare, but its not forbidden. 

So I added some (admittedly convoluted) logic to index the input effect and variance files based on the contrast weights. If you see a more elegant way to do it, feel free to suggest (but I'm trying to work with the contrast of making the `prepare_contrasts` function work for both level 1 and 2).

adelavega:
ping @effigies

fitlins/interfaces/abstract.py
```diff
@@ -56,7 +56,6 @@ class SecondLevelEstimatorInputSpec(TraitedSpec):
     contrast_info = traits.List(traits.Dict, mandatory=True)
     smoothing_fwhm = traits.Float(desc='Full-width half max (FWHM) in mm for smoothing in mask')
```
effigies:
```suggestion


```

fitlins/interfaces/nistats.py
```diff
@@ -196,46 +196,77 @@ def _run_interface(self, runtime):
...
 
-        # Dummy code contrast of input effects
-        design_matrix = pd.get_dummies(names)
+        mat = pd.get_dummies(names)
+        contrasts = prepare_contrasts(self.inputs.contrast_info, mat.columns)
 
-        # Fit single model for all inputs
-        model.fit(filtered_effects, design_matrix=design_matrix)
+        # Only fit model if any non-FEMA contrasts at this level
+        if any([c[2] != 'FEMA' for c in contrasts]):
```
```suggestion
        if any(c[2] != 'FEMA' for c in contrasts):
```

fitlins/interfaces/nistats.py
```diff
@@ -196,46 +196,77 @@ def _run_interface(self, runtime):
...
+            if len(filtered_effects) < 2:
+                raise RuntimeError(
+                    "At least two inputs are required for a 't' for 'F' "
+                    "second level contrast")
+            else:
```
effigies: No need for an `else` block when the `if` is for an exception.

```suggestion
```

fitlins/interfaces/nistats.py
```diff
@@ -196,46 +196,77 @@ def _run_interface(self, runtime):
...
+                model = level2.SecondLevelModel(smoothing_fwhm=smoothing_fwhm)
```
```suggestion
            model = level2.SecondLevelModel(smoothing_fwhm=smoothing_fwhm)
```

fitlins/interfaces/nistats.py
```diff
@@ -196,46 +196,77 @@ def _run_interface(self, runtime):
...
+                model.fit(filtered_effects, design_matrix=mat)
```
```suggestion
            model.fit(filtered_effects, design_matrix=mat)
```

fitlins/interfaces/nistats.py
```diff
@@ -196,46 +196,77 @@ def _run_interface(self, runtime):
...
                  **out_ents})
 
-            maps = model.compute_contrast(
-                second_level_contrast=weights,
-                second_level_stat_type=contrast_type,
-                output_type='all')
+            # Pass-through happens automatically as it can handle 1 input
+            if contrast_type == 'FEMA':
+                # Filter effects and variances based on weights
+                ix = mat.iloc[:, weights[0].astype('bool')].sum(axis=1)
```
effigies: I played with this, and I'm not sure it does what you want:

```Python
>>> import numpy as np
>>> import pandas as pd
>>> mat = pd.get_dummies(['a', 'b', 'c'])
>>> effects = ['d', 'e', 'f']
>>> ix = mat.iloc[:, np.array([1, 0, 1], dtype=bool)].sum(axis=1)
>>> ix
0    1
1    0
2    1
dtype: int64
>>> np.array(effects)[ix]                                                                      
array(['e', 'd', 'e'], dtype='<U1')
```

I think what you want is:

```Python
>>> np.array(effects)[ix]                                                                      
array(['d', 'f'], dtype='<U1')
```

So:

```suggestion
                ix = mat.iloc[:, weights[0].astype(bool)].sum(axis=1).astype(bool)
```

But isn't `mat` just an identity matrix? So isn't this doubling up?

```suggestion
                ix = weights[0].astype(bool)
```

Not sure if I'm missing something here, though...

fitlins/interfaces/utils.py
```diff
@@ -72,12 +74,14 @@ def _run_interface(self, runtime):
...
             val = getattr(self.inputs, key)
-            if len(val) != n:
-                raise ValueError(f"List lengths must match metadata. Failing list: {key}")
-            for md, obj in zip(orig_metadata, val):
-                metadata = md.copy()
-                metadata.update(md_map.get(key, {}))
-                self._results['metadata'].append(metadata)
-                self._results['out'].append(obj)
+            # Allow for missing values
+            if isdefined(val):
```
effigies: [`len(Undefined) == 0`](https://github.com/nipy/nipype/blob/63e5ef25eeaf8717ef271cc3397b41c2e754c5ef/nipype/interfaces/base/traits_extension.py#L67-L76)

fitlins/workflows/base.py
```diff
@@ -290,7 +290,6 @@ def _deindex(tsv):
                 SecondLevelModel(),
                 iterfield=['contrast_info'],
                 name='{}_model'.format(level))
```
```suggestion
                name='{}_model'.format(level))

```

fitlins/interfaces/utils.py
adelavega: huh?

effigies: Oh, nevermind.

fitlins/interfaces/nistats.py
adelavega: Yes, what I want is to get the column that are non-zero in a contrast. 

I think i am maybe doubling up let me think about it, but I'm trying to leave the design matrix the same across both types of contrasts.

adelavega: I think your suggestion might work though... For some reason I remember the `sum` being important but maybe not.

adelavega:
What happened to the `ignore` argument in the `LoadBIDSModel` input spec?

```
   load_entry_point('neuroscout-cli', 'console_scripts', 'neuroscout')()
  File "/src/neuroscout/neuroscout_cli/cli.py", line 61, in main
    command(deepcopy(args)).run()
  File "/src/neuroscout/neuroscout_cli/commands/run.py", line 57, in run
    retcode = run_fitlins(fitlins_args)
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/fitlins/cli/run.py", line 236, in run_fitlins
    smoothing=opts.smoothing, drop_missing=opts.drop_missing,
  File "/opt/miniconda-latest/envs/neuro/lib/python3.6/site-packages/fitlins/workflows/base.py", line 43, in init_fitlins_wf
    loader.inputs.ignore = ignore
traits.trait_errors.TraitError: Cannot set the undefined 'ignore' attribute of a 'LoadBIDSModelInputSpec' object.
```

effigies:
It's now part of the initial layout build.

adelavega:
Okay, tested this with #202 merged in, and my model is running! 

This is ready to merge as far as I'm concerned.

requirements.txt
```diff
@@ -1 +1 @@
-nistats @ git+https://github.com/nistats/nistats.git@b69e127020df05aa8ba540f9a8d799e1dad3d602
+nistats @ git+https://github.com/bthirion/nistats.git@explicit_fixed_effects#egg=nistats
```
effigies:
```suggestion
nistats @ git+https://github.com/nistats/nistats.git@e88a5f
```

effigies:
Can't commit for you. Feel free to merge once the dependency specifications are made consistent.

adelavega:
Will do. I'm gonna have to bug to do another release soon though!

PR #195, opened 2019-10-30
[WP] Add F-tests

### LINKED ISSUE #194: Add F-test contrasts
Looks like nistats supports F-tests. So let's add them to fitlins.

These would be particularly useful for models testing embedding models in NeuroScout. To do an omnibus test across all embeddings.

adelavega:
Closes #194 

Only added an example model so far

### DIFF
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -191,17 +191,15 @@ class SecondLevelModel(NistatsBaseInterface, SecondLevelEstimatorInterface, Simp
             if _match(out_ents, m):
                 filtered_effects.append(eff)
                 names.append(m['contrast'])
-        cnames = set(names)
 
-        # Dummy code condition of input effects
+        # Dummy code contrast of input effects
+        design_matrix = pd.get_dummies(names)
+
         # Fit single model for all inputs
-        design_matrix = pd.DataFrame(
-            {n: [1 if c == n else 0 for c in names] for n in cnames})
-
         model.fit(filtered_effects, design_matrix=design_matrix)
 
         for name, weights, contrast_type in prepare_contrasts(
-          self.inputs.contrast_info, cnames):
+          self.inputs.contrast_info, design_matrix.columns.to_list()):
             contrast_metadata.append(
                 {'contrast': name,
                  'stat': contrast_type,
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -206,8 +206,7 @@ class SecondLevelModel(NistatsBaseInterface, SecondLevelEstimatorInterface, Simp
                  **out_ents})
 
             maps = model.compute_contrast(
-                con_val=weights,
-                contrast_type=contrast_type,
+                second_level_contrast=weights,
                 second_level_stat_type=contrast_type,
                 output_type='all')
 
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -65,11 +65,11 @@ class DesignMatrix(NistatsBaseInterface, DesignMatrixInterface, SimpleInterface)
                 # Remove columns with NaNs
                 dense = dense[dense.columns[missing_columns == False]]
             elif missing_columns.any():
-                    missing_names = ', '.join(
-                        dense.columns[missing_columns].tolist())
-                    raise RuntimeError(
-                        f'The following columns are empty: {missing_names}. '
-                        'Use --drop-missing to drop before model fitting.')
+                missing_names = ', '.join(
+                    dense.columns[missing_columns].tolist())
+                raise RuntimeError(
+                    f'The following columns are empty: {missing_names}. '
+                    'Use --drop-missing to drop before model fitting.')
 
             column_names = dense.columns.tolist()
             drift_model = None if (('cosine00' in column_names) |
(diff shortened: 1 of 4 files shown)


adelavega:
As far as I can tell, F-tests should already be working at the first level (which is the only thing of any urgency to me).

At higher levels, it looks like we are modeling the intercept so that wont work.

effigies:
What does it mean for F-tests to be working? What are you looking at to assess this?

adelavega:
Correction: Just talked to Tal, and I think we've figured it out.

To do a proper F-test it needs to be done at the dataset level, not at the first-level. Doing one at the first level is going to do the same linear weighting as if you just did an equivalent t-test-- that is give you the average (although the stats themselves might be of interest to someone that is doing single subject stats)

So what that means is that at the second level, we need to create a design matrix which codes (as columns) the condition type of each input  (rows) across subjects.

In this ds003 example, there are 2 conditions and 3 subjects. That results in 6 effect images at the dataset level.

For t-tests, since its only 1 dimension, we simply use the intercept to indicate which input effects belong to the relevant condition. So for example for a t-test of words:

| file_name        | intercept | 
| ------------- |:-------------:| 
| sub1_word   | 1 |
| sub2_word   | 1 | 
| sub3_word   | 1 | 
| sub1_pseudoword   | 0 | 
| sub2_pseudoword   | 0 | 
| sub3_pseudoword   | 0 |

For f-tests, this won't work, since its 2+ dimensional. Instead, we need to code condition identity as columns in the design matrix:. E.g.:

| file_name        | word | pseudo-word  |
| ------------- |:-------------:| -----:|
| sub1_word   | 1 | 0 |
| sub2_word   | 1 | 0 |
| sub3_word   | 1 | 0 |
| sub1_pseudoword   | 0 | 1 |
| sub2_pseudoword   | 0 | 1 |
| sub3_pseudoword   | 0 | 1 |

On this design matrix you could then compute one-sample t-tests:

`model.compute_contrast('word', type='t')`
or:
`mode.compute_contrast([1, 0], type='t')`

or F-tests (here's an omnibus)

`model.compute_contrast([[1, 0], [0, 1]], type='F')`

I'll update this PR with these changes.

adelavega:
@effigies what I mean is that f-tests already can be passed in at the first level. But they don't really make sense to do in most scenarios, as we want F-tests at the dataset level.

adelavega:
I think this will work, but it's kind of ugly. I'm basically creating the dummy coded design matrices that I put above manually, from the unique condition names that are passed in as `stat_metadata`.

It seems like we should be using pybids more intelligently.

`analysis.steps[-1].get_design_matrix`

But when I look at the output of that, the only column that is relevant is `condition` (which I'd then dummy code), and I can get from `names`.

The other minor issue is that now the weights are created based on `set(names)` because the weights of the contrast are now relative to the design matrix column, nor rows (like before). But for #191 the "weights" for fixed-effects are still rows, so we'll have to integrate that.

adelavega:
Okay cleaned that up a bit.

effigies:
Fixed the docs and packaging tests on `master`. I'll look into the F-tests locally now.

fitlins/interfaces/nistats.py
```diff
@@ -184,37 +184,30 @@ def _run_interface(self, runtime):
...
         input_effects = _flatten(self.inputs.effect_maps)
-        # XXX nistats should begin supporting mixed effects models soon
-        # input_variances = _flatten(self.inputs.variance_maps)
-        input_variances = [None] * len(input_effects)
 
         filtered_effects = []
-        filtered_variances = []
         names = []
-        for m, eff, var in zip(stat_metadata, input_effects, input_variances):
+        for m, eff, var in zip(stat_metadata, input_effects):
```
```suggestion
        for m, eff in zip(stat_metadata, input_effects):
```

fitlins/interfaces/nistats.py
```diff
@@ -184,37 +184,30 @@ def _run_interface(self, runtime):
...
-            design_matrix = pd.DataFrame(
-                {'intercept': weights[weights != 0]})
-
-            model.fit(effects, design_matrix=design_matrix)
-
             maps = model.compute_contrast(
+                con_val=weights,
+                contrast_type=contrast_type,
                 second_level_stat_type=contrast_type,
                 output_type='all')
```
effigies: These aren't valid arguments:

https://github.com/nistats/nistats/blob/4b7adc71416504720fefcaf532ac16e7a13ae3f0/nistats/second_level_model.py#L413-L415

adelavega: Oh, I guess the args differ from the standalone function.

I wonder how you pass a 2d contrast matrix then? Anyways I'm out the rest of the week so I'll look at it Monday

adelavega: Oh, I guess the args differ from the standalone function.

I wonder how you pass a 2d contrast matrix then? Anyways I'm out the rest of the week so I'll look at it Monday

effigies:
Fixed some issues, updated the CI to use the example model from the current branch, so now we can inspect the outputs in the artifacts.

Example: https://1571-108019675-gh.circle-artifacts.com/0/tmp/ds003/derivatives/fitlins/reports/model-ds003Model001.html

effigies:
Rebased, so you should `git rebase origin/enh/f-tests` (or `git reset --hard origin/enh/f-tests`) before continuing.

effigies:
Nvm. This is a useful improvement already, and we can do more PRs.

PR #200, opened 2019-11-22
WIP: Circle test for AFNI

effigies:
Extending on #171 to run tests.

Diagnosing issues led to #198 and #199. Assuming those pass without issue, I'll rebase to refocus this PR.

@leej3


effigies:
@leej Have you tested this locally? At least on Circle, it seems to take a *very* long time on the post-RemlFit phase of FirstLevelModel.

leej3:
Ah. That would be a bug. I think a change to 3dpval makes a certain version
of AFNI (round about when I submitted the pull request) a requirement. A
version check will probably have to be added to defensively fail. I think
3dPval hangs with your current version.

leej3:
A new neurodebian release shouldn't be too long now. There is one pernicious bug that has slowed down the process: https://github.com/afni/afni/issues/131

effigies:
Sounds good. Is it important to use 3dPval? We can also do simple math in Python.

leej3:
It’s not no. I was just doing a pure afni implementation to have the comparison available. And nistats compute all of the stat maps at once so it seemed silly to use that...

effigies:
@leej3 Any news regarding AFNI in Neurodebian?

leej3:
I'm still working through it. I solved the aforementioned bug but I still have a bit more to do. I am struggling to get suma to work in my test container. The OpenGL dependency makes things tricky. I'm getting through it, it's just painfully slow. If it takes too much longer I can rethink things: I can either alter my CI tests to not use a container for the tests on CI (though I may very well run into the same issues because I will need headless execution), or skip the packaging of suma for now and just do a release without suma. 

Here's hoping I just figure it out quickly.

effigies:
Okay. Looks like we're getting somewhere, but we're also hanging (working?) for hours. I'll try to see if I can get this going locally so I can debug.

leej3:
I think the 'aforementioned bug' was ambiguous. It was referring to an issue with linking in suma. The bug with 3dPval mentioned in this thread is still going to cause the tests to hang. They won't work until a more recent neurodebian release is made. Still working on that. There is endless list of details to work through but its moving forward.

effigies:
Oh, I see. I thought 3dPval was just missing...

Let me know if there's something useful I can do on the AFNI/NeuroDebian end.

leej3:
Thanks, will do.

Shotgunosine:
Closing since these changes were cherrypicked into #171 .

PR #204, opened 2019-12-11
Add desc=preproc as filter when finding preproc BOLD files

adelavega:
Related issue: https://github.com/bids-standard/pybids/issues/557

Stems from having `validate=False`, which we only used to do when necessary, but now is applied across the whole master Layout

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -398,13 +398,13 @@ class BIDSSelect(SimpleInterface):
 
             if len(bold_file) == 0:
                 raise FileNotFoundError(
-                    "Could not find BOLD file with entities {}"
-                    "".format(selectors))
+                    "Could not find BOLD file in {} with entities {}"
+                    "".format(layout.root, selectors))
             elif len(bold_file) > 1:
                 raise ValueError(
-                    "Non-unique BOLD file with entities {}.\n"
+                    "Non-unique BOLD file in {} with entities {}.\n"
                     "Matches:\n\t{}"
-                    "".format(selectors,
+                    "".format(layout.root, selectors,
                               "\n\t".join(
                                   '{} ({})'.format(
                                       f.path,
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -398,13 +398,13 @@ class BIDSSelect(SimpleInterface):
 
             if len(bold_file) == 0:
                 raise FileNotFoundError(
-                    "Could not find BOLD file in {} with entities {}"
-                    "".format(self.inputs.bids_dir, selectors))
+                    "Could not find BOLD file with entities {}"
+                    "".format(selectors))
             elif len(bold_file) > 1:
                 raise ValueError(
-                    "Non-unique BOLD file in {} with entities {}.\n"
+                    "Non-unique BOLD file with entities {}.\n"
                     "Matches:\n\t{}"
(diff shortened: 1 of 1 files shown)


adelavega:
Uhh, let me try that again, I tried to rebase and failed.

adelavega:
Tested, and neuroscout-cli models are running with this change.

Not sure why this didn't come up before, but I think I hadn't rebuilt my fitlins image in a while, and maybe it was on a previous pybids version or something. Upgrading to 0.6.0 did it though.

I guess bugfix release will be needed, ugh.

effigies:
I suspect it's because I only used `validate=False` when constructing models. Most of the layouts I constructed were validated.

fitlins/interfaces/bids.py
```diff
@@ -231,7 +231,7 @@ def _load_level1(self, runtime, analysis):
             TR = ents.pop('RepetitionTime', None)  # Required field in seconds
             if TR is None:  # But is unreliable (for now?)
                 preproc_files = analysis.layout.get(
-                    extension=['.nii', '.nii.gz'], **ents)
+                    extension=['.nii', '.nii.gz'], desc='preproc', **ents)
```
effigies: Is there any chance `ents` could already contain `'desc'`? If so, this is setting us up for an error.

fitlins/interfaces/bids.py
```diff
@@ -393,18 +393,18 @@ def _run_interface(self, runtime):
...
         entities = []
         for ents in self.inputs.entities:
-            selectors = {**self.inputs.selectors, **ents}
+            selectors = {**self.inputs.selectors, 'desc': 'preproc', **ents}
             bold_file = layout.get(**selectors)
 
             if len(bold_file) == 0:
                 raise FileNotFoundError(
-                    "Could not find BOLD file in {} with entities {}"
-                    "".format(self.inputs.bids_dir, selectors))
```
effigies: What about `layout.root`?

fitlins/interfaces/bids.py
```diff
@@ -393,18 +393,18 @@ def _run_interface(self, runtime):
         mask_files = []
         entities = []
         for ents in self.inputs.entities:
-            selectors = {**self.inputs.selectors, **ents}
+            selectors = {**self.inputs.selectors, 'desc': 'preproc', **ents}
```
effigies: By putting it first, if `'desc'` is in `selectors` or `ents`, it will override.

```suggestion
            selectors = {'desc': 'preproc', **self.inputs.selectors, **ents}
```

Unless your goal was specifically to override selectors and be overridden by ents?

fitlins/interfaces/bids.py
adelavega: It doesn't

fitlins/interfaces/bids.py
adelavega: Nope, that works.

adelavega:
> I suspect it's because I only used `validate=False` when constructing models. Most of the layouts I constructed were validated.

Yep, that's exactly what happened.

adelavega:
I'm not too happy about `validate=False` for other reasons though. So maybe we should keep this in mind for a future fix/alternate way of doing things.

effigies:
What if we enable validation but package our own config that handles `smdl` suffixes and `model` entities?

adelavega:
That's probably the best thing to do yeah

adelavega:
Merging, hope that's allright.

PR #210, opened 2020-01-15
RF: Accommodate and use new pybids model specification

effigies:
* [x] Restore functionality after pinning to bids-standard/pybids#548
* [ ] Switch to new model specification

cc @tyarkoni


effigies:
This got out-of-date. It's being included in #239.

PR #214, opened 2020-01-30
FIX: Validate BIDS datasets except when searching for models

### LINKED ISSUE #206: Validate BIDSLayout
Currently, `BIDSLayout` is not validated, because model files are not valid BIDS (see: https://github.com/poldracklab/fitlins/pull/204). 

This can cause other unintended problems, so it would be best to add a custom config to pass initi pybids for these files that don't pass validation.

effigies:
Fixes #206.

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -210,7 +210,6 @@ def run_fitlins(argv=None):
     layout = BIDSLayout(opts.bids_dir,
                         derivatives=derivatives,
                         ignore=ignore,
-                        validate=False,
                         force_index=force_index,
                         database_path=database_path,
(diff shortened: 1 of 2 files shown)


effigies:
I think this should be reasonably cheap, as it only comes into play if not explicitly passing a model, and skipping `sub-*` and `.git/` has a noticeable speedup on runtime.

adelavega:
Nice.

adelavega:
I guess one comment I have though is that it might be a reasonable CLI option whether or not to validate the `BIDSLayout`.

effigies:
Not sure what that would buy a user.

PR #216, opened 2020-02-26
MNT: Bump pybids to 0.10.2

adelavega:
- Specifically, this should fix the Replace transformation


adelavega:
Looks like its failing because pypi hadn't updated

effigies:
Yup. Rerunning.

adelavega:
no rush on this btw, im mapping the latest pybids on my docker image

PR #217, opened 2020-02-26
WIP: Add Diagnostics module

### LINKED ISSUE #213: Design Matrix diagnostic module
It would be useful to have a module of diagnostic checks on design matrices, like other analysis software provides.

For example, things such as checking for rank-deficiency of the design-matrix.

This is something I planned to add to neuroscout, but it makes more sense to put it somewhere generalization.

Since pybids will now only output a `df` and not a `design_matrix` per se, @tyarkoni suggested putting this module here.

The idea is to have a `Diagnostic` class, with something like a `run` function that takes a design matrix. The class will also have many functions starting with `check_` and the `run` function will apply all of them. We could have on/off flags in the `run` function to disbale some checks, or simply just return a long list of warnings that the user can feel free to ignore.

These can then be rendered on Neuroscout's frontend by importing from fitlins, as well as rendered by fitlin reports.

adelavega:
Closes #213 

Just a skeleton for now, but let me know what you think of the structure.

One point for the debate is the plots. If we put them in there (which would be nice IMO to centralize efforts), I need them to be in altair. 

This would actually be pretty nice for the web reports as they are natively JS plots, but would require some refactoring.

### DIFF
--- a/fitlins/viz/diagnostics.py
+++ b/fitlins/viz/diagnostics.py
@@ -1,9 +1,25 @@
 class Diagnostics:
     """ Warnings and plots generated from a first-level design matrix """
 
+    def __init__(self):
+        all_methods = [
+            func for func in dir(self) if callable(getattr(self, func))]
+        self._all_checks = [
+            func for func in all_methods if func.startswith('_check')]
+        self._all_plots = [
+            func for func in all_methods if func.startswith('_plot')]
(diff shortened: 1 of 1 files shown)


adelavega:
This is stale so I'm closing. I'll re-open if I start working on this again

PR #224, opened 2020-03-26
ENH: Save model level output (r-squared)

### LINKED ISSUE #221: Output model level outputs
Such as log-likelihood, residuals, etc...

These can all be grabbed from nistats model object after its been fit, and output in subject directories as niftis.

adelavega:
- Saves r-squared from model level outputs. 

Closes #221


adelavega:
@tyarkoni are `r_squared` and `residuals` maps what you for the BERT analysis?

Those are the two maps exposed by `FirstLevelModel`

adelavega:
hey @jdkent, you mind taking a look? should be fairly straight forward, but looks like you contributed adding r-squared and residuals to the `FirstLevelModel` object to nistats, so let me know what you think!

adelavega:
Looks like tests are failing because of a MemoryError!

Unfortunately, the new images are surprisingly memory intensive to produce.

fitlins/interfaces/nistats.py
```diff
@@ -122,18 +122,33 @@ def _run_interface(self, runtime):
         smoothing_fwhm = self.inputs.smoothing_fwhm
         if not isdefined(smoothing_fwhm):
             smoothing_fwhm = None
+
         flm = level1.FirstLevelModel(
+            minimize_memory=False,
```
effigies: I guess it's this line that's killing us?

adelavega: Yes and no. 

Yes because that argument must be set to true for those images to be saved to that object. Otherwise, because they are not super necessary they are deleted along with a bunch of other stuff.

@jdkent maybe there's another way to access these images without triggering a bunch of other stuff to be kept around?

adelavega: From what I can tell, that argument saves the `SimpleRegressionResult` object, and without it, you can't get the other model outputs. 

nistats is also archived, so proposing an alternative for now is not possible.

Maybe we can just make this optional in fitlins? I'll raise an issue over there and see what they say.

adelavega:
When I test locally, my docker container only hits around 400mb of memory usage. Looks like Circle CI's memory is probably quite low, and creating the residuals map (which is 4D) is momentarily very memory intensive. 

I don't actually need the residuals, just the r-squared, so removing for now to see if tests pass. We could make some or all of these optional by user CLI flags.

jdkent:
popping in to say that I do not know of a currently builtin way to save only what we want, and delete other pieces of information, sorry! (working on filtering my notifications so I can see mentions sooner). 

I'll take a look and see if I have anything helpful to add.

adelavega:
Thanks @jdkent 

Well, the important tests are passing now! So I think just computing the residuals is the most memory intensive part. For now, I'll just output r-squared as that's what I care about, but we could add a CLI option for the others.

adelavega:
@effigies this is what the output file looks like. look reasonable? it's at the run-level for the whole model:

`sub-01_task-SherlockMovie_space-MNI152NLin2009cAsym_stat-rSquare_statmap.nii.gz`

jdkent:
How you are getting `r_square` from nistats first level outputs looks reasonable to me.

To fix the memory issues, a more flexible interface within `FirstLevelModel` is needed. (Perhaps having users specify what model outputs they want when they call `fit`).

Similar to how effigies suggested the ['all' output for computing contrasts](https://github.com/nilearn/nilearn/blob/318bb5a31c1fb9c9afc737ef1f7b7e48c8ace334/nilearn/stats/first_level_model/first_level_model.py#L556-L558), I could see a parameter that saves `r_square`, `residuals`, `predicted`, or `all`.

effigies:
Either is fine with me.

adelavega:
Looks like there's an issue with a duecredit dependency: https://github.com/brechtm/citeproc-py/issues/94

May want to remove that for the time being.

adelavega:
hey @jdkent do you know if I can access other model attributes such as `logL`? Is there any reason that is not available at the `FirstModelResult` level?

I'm going to open another PR trying to get that file to be output as well. 

@effigies Do you think these should always be output, or by CLI option? I can work that into the next PR as well.

jdkent:
you should be able to access LogL with the private method `_get_voxelwise_model_attribute`
like `flm._get_voxelwise_model_attribute("LogL", result_as_time_series=False)`

(LogL is just a 3-d output right?)

adelavega:
yep, thanks! looked at your code and figured that out.

PR #228, opened 2020-04-06
Temporarily remove duecredit until dependency issue is resolved

adelavega:


yarikoptic:
If you could tolerate a few hours wait, I might mitigate at citeproc level of not resolved properly by rnc2rng

effigies:
Looks like it's resolved.

PR #230, opened 2020-04-06
Add logL as model output

adelavega:
- Also output log-likelkihood (logL) from model

### DIFF
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -134,17 +134,50 @@ class FirstLevelModel(NistatsBaseInterface, FirstLevelEstimatorInterface, Simple
         # Save model level images
         model_maps = []
         model_metadata = []
-        for param in ['r_square', 'logL']:
-            model_metadata.append(
-                {'stat': param,
-                 **out_ents}
-                )
-            fname = fname_fmt('model', param)
-            output = flm._get_voxelwise_model_attribute(
-                param, result_as_time_series=False)[0]
 
-            output.to_filename(fname)
-            model_maps.append(fname)
+        # R-square
+        model_metadata.append(
+            {'stat': 'r_square',
+             **out_ents}
+            )
+        fname = fname_fmt('model', 'r_square')
+        flm.r_square()[0].to_filename(fname)
+        model_maps.append(fname)
+
+        # Manually compute log-likelihood
+
+        def _get_voxelwise_logL(flm):
+            """
+            Returns
+            -------
+            output : list
+                a list of Nifti1Image(s)
+            """
+
+            output = []
+            for design_matrix, labels, results in zip(self.design_matrices_,
+                                                      self.labels_,
+                                                      self.results_
+                                                      ):
+
+                voxelwise_attribute = np.zeros((1, len(labels)))
+
+                for label_ in results:
+                    label_mask = labels == label_
+                    voxelwise_attribute[:, label_mask] = getattr(results[label_],
+                                                                 'logL')
+
+                output.append(self.masker_.inverse_transform(voxelwise_attribute))
+
+                return output
+
+        model_metadata.append(
+            {'stat': 'logL',
+             **out_ents}
+            )
+        fname = fname_fmt('model', 'logL')
+        _get_voxelwise_logL(flm)[0].to_filename(fname)
+        model_maps.append(fname)
 
         effect_maps = []
         variance_maps = []
--- a/fitlins/interfaces/nistats.py
+++ b/fitlins/interfaces/nistats.py
@@ -38,6 +38,33 @@ def prepare_contrasts(contrasts, all_regressors):
     return out_contrasts
 
 
+def _get_voxelwise_logL(flm):
+    """
+    Given a FirstLevelModel, computed voxel wise log-likelihood map
+    Returns
+    -------
+    output : list
+        a list of Nifti1Image(s)
+    """
+
+    output = []
+    for design_matrix, labels, results in zip(flm.design_matrices_,
+                                              flm.labels_,
+                                              flm.results_
+                                              ):
+
+        voxelwise_attribute = np.zeros((1, len(labels)))
+
+        for label_ in results:
+            label_mask = labels == label_
+            voxelwise_attribute[:, label_mask] = getattr(results[label_],
+                                                         'logL')
+
+        output.append(flm.masker_.inverse_transform(voxelwise_attribute))
+
+    return output
+
+
 class DesignMatrix(NistatsBaseInterface, DesignMatrixInterface, SimpleInterface):
 
     def _run_interface(self, runtime):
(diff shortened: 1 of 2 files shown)


adelavega:
@effigies @jdkent I'm testing out adding `log-likelihood` to fitlins outputs.

Unfortunately, it was actually not available with `_get_voxelwise_model_attribute` because `logL` is not an attribute of `SimpleRegressionResult`, and that's what's being checked when it validates the arguments to the function (it shouldbe checking `RegressionResult` which is the type of object returned when `minimize_memory` is True). It's also not an attribute, but a function of the result object. 

nistats is archived so I can't get that fix upstream, so that custom function to retrieve it would have to live here for now.

adelavega:
Okay, great the maps look good. 

Only other point is what they key should be. `logL` would reflect nistats, but we could go with `logLikelihood` for more clarity.

jdkent:
Thanks for investigating this @adelavega, and sorry it was more pain to implement than anticipated.

> Unfortunately, it was actually not available with _get_voxelwise_model_attribute because logL is not an attribute of SimpleRegressionResult

I don't think `_get_voxelwise_model_attribute` would work at all with `SimpleRegressionResult` since the function [raises an error](https://github.com/nilearn/nistats/blob/bd4e27fdd5487b1ff920d4ae2473a7671153d2b3/nistats/first_level_model.py#L646) if `minimize_memory` is True (which mean that the model is using [`SimpleRegressionResult` instead of `RegressionResult`](https://github.com/nilearn/nistats/blob/bd4e27fdd5487b1ff920d4ae2473a7671153d2b3/nistats/first_level_model.py#L518-L521) to store output).

Your point still stands that `logL` is also not an attribute directly from `RegressionResults`:
```
>>> dict(vars(RegressionResults)).keys()
dict_keys(['__module__', '__doc__', '__init__', 'wdesign', 'wY', 'wresid', 'resid', 'residuals', 'norm_resid', 'normalized_residuals', 'predicted', 'SSE', 'r_square', 'MSE'])
```

But `logL` is an attribute from [`LikelihoodModelResults`](https://github.com/nilearn/nistats/blob/bd4e27fdd5487b1ff920d4ae2473a7671153d2b3/nistats/model.py#L20) (which `RegressionResults` inherits from)

```
>>> dict(vars(LikelihoodModelResults)).keys()
dict_keys(['__module__', '__doc__', '__init__', 'df_resid', 'logL', 't', 'vcov', 'Tcontrast', 'Fcontrast', 'conf_int', '__dict__', '__weakref__'])
```

I did not do a diligent check to make sure the attributes from the inherited class (`LikelihoodModelResults`) were passed into the inheriting class (`RegressionResult`), likely because of how `@setattr_on_read` works.

> It's also not an attribute, but a function of the result object.

`logL` is a function for `SimpleRegressionResults`, but [`logL` should be an attribute for `RegressionResults`](https://github.com/nilearn/nistats/blob/bd4e27fdd5487b1ff920d4ae2473a7671153d2b3/nistats/model.py#L89-L94) (once calculated), unless if my understanding of [`@setattr_on_read`](https://nipy.org/nibabel/reference/nibabel.onetime.html#setattr-on-read) is wrong.

So I would change:
[`all_attributes = dict(vars(RegressionResults)).keys()`](https://github.com/nilearn/nistats/blob/bd4e27fdd5487b1ff920d4ae2473a7671153d2b3/nistats/first_level_model.py#L635) 
to
```
all_attributes = set(dict(vars(LikelihoodModelResults)).keys()).union(dict(vars(RegressionResults)).keys())
```
So any attributes that `RegressionResults` directly has or inherits from would be valid.


Finally, I also like `logLikelihood` for more clarity.

fitlins/interfaces/nistats.py
```diff
@@ -134,14 +134,50 @@ def _run_interface(self, runtime):
...
+            {'stat': 'r_square',
+             **out_ents}
+            )
+        fname = fname_fmt('model', 'r_square')
+        flm.r_square[0].to_filename(fname)
+        model_maps.append(fname)
+
+        # Manually compute log-likelihood
+
+        def _get_voxelwise_logL(flm):
```
jdkent, line 149: I don't know if it would be easier, but you could also patch the `FirstLevelModel` class with a more correct `_get_voxelwise_model_attribute` function definition, and use [`FirstLevelModelPatch` here](https://github.com/poldracklab/fitlins/pull/230/files#diff-efda2d1eedc98c2ee14427f38cf64710L126).
The main benefit being if this change is pulled into nilearn/nistats, then you can just delete the patched class and use the "normal" one again and not potentially change how you are solving the problem (by defining a custom function `_get_voxelwise_logL`).

Overthinking this since your solution works and is also good (and ultimately fewer lines of code).


```
from nistats.regression import RegressionResults
from nistats.model import LikelihoodModelResults

class FirstLevelModelPatch(FirstLevelModel)
    def _get_voxelwise_model_attribute(self, attribute,
                                       result_as_time_series):
        """Transform RegressionResults instances within a dictionary
        (whose keys represent the autoregressive coefficient under the 'ar1'
        noise model or only 0.0 under 'ols' noise_model and values are the
        RegressionResults instances) into input nifti space.
        Parameters
        ----------
        attribute : str
            an attribute of a RegressionResults instance.
            possible values include: resid, norm_resid, predicted,
            SSE, r_square, MSE.
        result_as_time_series : bool
            whether the RegressionResult attribute has a value
            per timepoint of the input nifti image.
        Returns
        -------
        output : list
            a list of Nifti1Image(s)
        """
        # check if valid attribute is being accessed.
         # CHANGING THIS LINE TO BE BETTER
        all_attributes = set(dict(vars(LikelihoodModelResults)).keys()).union(dict(vars(RegressionResults)).keys())
        possible_attributes = [prop
                               for prop in all_attributes
                               if '__' not in prop
                               ]
        if attribute not in possible_attributes:
            msg = ("attribute must be one of: "
                   "{attr}".format(attr=possible_attributes)
                   )
            raise ValueError(msg)

        if self.minimize_memory:
            raise ValueError(
                'To access voxelwise attributes like '
                'R-squared, residuals, and predictions, '
                'the `FirstLevelModel`-object needs to store '
                'there attributes. '
                'To do so, set `minimize_memory` to `False` '
                'when initializing the `FirstLevelModel`-object.')

        if self.labels_ is None or self.results_ is None:
            raise ValueError('The model has not been fit yet')

        output = []

        for design_matrix, labels, results in zip(self.design_matrices_,
                                                  self.labels_,
                                                  self.results_
                                                  ):
            if result_as_time_series:
                voxelwise_attribute = np.zeros((design_matrix.shape[0],
                                                len(labels))
                                               )
            else:
                voxelwise_attribute = np.zeros((1, len(labels)))

            for label_ in results:
                label_mask = labels == label_
                voxelwise_attribute[:, label_mask] = getattr(results[label_],
                                                             attribute)

            output.append(self.masker_.inverse_transform(voxelwise_attribute))

            return output
```

effigies: I think subclassing things from nistats is probably going to be more effort than it's worth at this point, but worth thinking about making these contributions upstream to nilearn post-0.7.0.

@adelavega Can you define this function at the file-level, rather than nested inside this scope?

adelavega: Yeah, I can do that. I think that's easiest for now.

jdkent: Sounds good to me, I can reference an issue in nilearn.

adelavega:
Looks like tests need to be re-run, random installation failure

effigies:
Tests passing. Happy with this?

adelavega:
Let me test locally once more.

adelavega:
Okay, LGTM, so feel free to merge once tests pass.

adelavega:
Actually let's hold off, @tyarkoni is checking out if this is the log-likelihood we want.

adelavega:
OK It looks good

PR #237, opened 2020-06-18
WIP: Begin converting BIDS interfaces to Pydra tasks

effigies:
This PR shows two initial attempts at replacing the `ModelSpecLoader` interface with Pydra `Task`s. The first subclasses `TaskBase`, the second uses `@pydra.mark.task` and function annotation.

This has overall been delayed by my partially-retrieved test datasets, which have caused issues first with Pydra (hashing a directory with dangling symlinks) and then PyBIDS (loading metadata from dangling symlinks).

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -6,6 +6,111 @@ import json
 import shutil
 import numpy as np
 import nibabel as nb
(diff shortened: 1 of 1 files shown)


effigies:
cc @satra

PR #239, opened 2020-06-21
ENH: Add CIFTI-2 support

effigies:
This builds on @satra's PR (#126).

It also supersedes #210, although that could be done separately. We should probably release PyBIDS 0.11 once this is working with `master`.

Still some things to be done, but this gets through run-level models on my test data.

TODO:

* [ ] Set up tests with ds003
* [x] Test second-level
* [x] Fix/skip plotting for CIFTI-2
* [x] Figure out model maps with CIFTI-2
* Probably some other bits I haven't thought of yet.


fitlins/interfaces/nistats.py
```diff
@@ -149,31 +174,29 @@ def _run_interface(self, runtime):
...
+        #     'log_likelihood': _get_voxelwise_logL(flm)[0]
+        # }
+
+        # model_maps = []
+        # model_metadata = []
+        # for attr, img in model_attr.items():
+        #     model_metadata.append{'stat': attr, **out_ents})
+        #     fname = fname_fmt('model', attr)
+        #     img.to_filename(fname)
+        #     model_maps.append(fname)
```
effigies, line 199: @adelavega @jdkent I had to filter this out for the moment because CIFTI-2s can't be fit with a `FirstLevelModel`. If either of you has some time to look at how to restore these with just the `run_glm` outputs, that would be great. I'll come back around to this after I get plotting working.

effigies:
@satra I'm currently able to get this to fit at the run level and do a fixed-effects combination at the subject level. The reporting needs work (subject level shows run 1) and the group level is crashing due to a failure to filter inputs properly, but feel free to give this a try and see if you're getting sensible results at the first/second level.

effigies:
The current version of ds003 that I'm testing on doesn't have usable CIFTI-2 files. I'm planning on re-running, but I'm not sure it's worth holding this up for that.

@adelavega Can you test neuroscout on this branch, or do I need to merge into master first?

bpinsard:
Thanks for that nice feature!
Does smoothing works on the CIFTI space, prior to computing the maps?

effigies:
No, I don't have smoothing implemented for CIFTI, yet.

PR #253, opened 2020-09-19
Only drop modulation if needed

adelavega:
FWIW I rebuilt the image and without this fix seemed to work so I think it's probably dependent on pandas versioning. 

Still, doesn't hurt to do this.


effigies:
Seems reasonable. I'll look at the test failures ASAP.

PR #262, opened 2020-11-16
[TST] Add dummy dataset for pytest/local testing

jdkent:
This pull request includes:
- a clumsy attempt to generate small images and other necessary files for "quick" local testing of code
  - (eventually could be useful for creating more unittests)
- a test of the entire workflow using the small images
- running pytest on circleci
- combining coverage from pytest and ds003


setup.cfg
```diff
@@ -33,7 +33,10 @@ install_requires =
 
 [options.extras_require]
 duecredit = duecredit
-test = coverage<5
+test =
+    coverage<5
```
effigies, line 37: Can we update coverage at this point? Would you mind making a new branch with this where we use latset?

jdkent:
I think I need help to determine how to properly `persist_to_workspace` and `attach_workspace`, what I thought should work does not appear to (in the upload_coverage job).

effigies:
I would just upload coverage separately in each job. Codecov is fine receiving multiple reports.

effigies:
Sorry, I should have noted that this was bound to conflict with the big AFNI PR. Care to merge/rebase master?

PR #264, opened 2020-11-19
[CI] Make faster?

jdkent:
I made CI take a long time in #262 , and now I'm going to try to fix it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,6 +261,8 @@ jobs:
           name: Run FitLins Database-file
           no_output_timeout: 2h
           command: |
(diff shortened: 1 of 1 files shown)


effigies:
I think what we probably want to do is split the DB tests from the basic tests, so that they're only run on one job, rather than parameterized across estimators. It was mostly done that way because there was only one real test.

effigies:
@jdkent Sorry for jumping the gun on the last one. Is this ready to go?

jdkent:
no worries, I should mark as work in progress until I'm ready, this is good to go!

effigies:
Awesome, thanks!

PR #268, opened 2020-12-07
MNT: Update PyBIDS and nilearn

### LINKED ISSUE #261: nilearn 0.7.0 contains nistats
[nilearn 0.7.0 was released earlier today](https://github.com/nilearn/nilearn/releases/tag/0.7.0) which is the new home for nistats, in [`nilearn.glm`](https://github.com/nilearn/nilearn/tree/0.7.0/nilearn/glm). The next time dependencies are updated, nistats can be dropped.

effigies:
Drop nistats, put some pybids warnings to rest.

Closes #261.

### DIFF
--- a/fitlins/cli/run.py
+++ b/fitlins/cli/run.py
@@ -19,12 +19,12 @@ from argparse import ArgumentParser
 from argparse import RawTextHelpFormatter
 from multiprocessing import cpu_count
 
-from bids.layout import BIDSLayout
+import bids
 from bids.analysis import auto_model, Analysis
 
 from .. import __version__
 from ..workflows import init_fitlins_wf
-from ..utils import bids, config
+from ..utils import bids as fub, config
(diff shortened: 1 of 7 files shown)


fitlins/cli/run.py
```diff
@@ -19,12 +19,12 @@
 from argparse import RawTextHelpFormatter
 from multiprocessing import cpu_count
 
-from bids.layout import BIDSLayout
+import bids
```
adelavega, line 22: why not import both object instead of whole library? nitpick though otherwise lgtm

adelavega:
awesome.

fitlins/cli/run.py
effigies: My aesthetic preferences go through phases... bids is a short namespace, so seems cheap to be explicit.

PR #275, opened 2021-01-21
ENH: modify path patterns to make task optional

adelavega:
It should not be required to specify a task in a model (i.e. run a model across all tasks)

I modified the `PATH_PATTERNS` to make task optional. 

Need to test a bit more.

### DIFF
--- a/fitlins/workflows/base.py
+++ b/fitlins/workflows/base.py
@@ -107,24 +107,24 @@ def init_fitlins_wf(database_path, out_dir, analysis_level, space,
 
     # Set up common patterns
     image_pattern = 'reports/[sub-{subject}/][ses-{session}/]figures/[run-{run}/]' \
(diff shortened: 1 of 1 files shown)


effigies:
Gah. Failed to read all the way to the bottom. Hope your tests went well.

adelavega:
Oops. I guess you have no objection in principle. I'll put WIP in the title next time. Tests still running.

adelavega:
That worked for me, now testing again a use case w/ a task.

By the way, the group level files will look something like this with no task:

`space-MNI152NLin2009cAsym_contrast-rms_stat-effect_statmap.nii.gz`

A bit ugly but I guess that's bids compliant?

effigies:
Yup.

adelavega:
Okay, yup. For the record this didn't cause any problems for models w/ task for me

effigies:
Yeah, it shouldn't have, since the only issue might be change in fallback behavior, and we're not sending multiple patterns to the same nodes.

PR #276, opened 2021-01-27
Make drift model configurable via CLI

### LINKED ISSUE #274: disable temporal filtering / drift model with CLI flag
Currently, if fitlins doesn't fit temporal filtering regressors ala fmirprep, it will enable them by default in nistats. 

I'm not sure if this should be the default behavior, as this filtering could have already been performed on the preprocessed inputs, and not require double filtering. Either way, this should be configurable behavior, I think.

adelavega:
Closes #274 

I debated if this should be a CLI flag, but honestly the CLI is already feeling a bit bloated, and I think automatically doing anything to do the design matrix is against the design philosopy of BIDS StatsModel / fitlins.

### DIFF
--- a/fitlins/workflows/base.py
+++ b/fitlins/workflows/base.py
@@ -5,7 +5,7 @@ import warnings
 def init_fitlins_wf(database_path, out_dir, analysis_level, space,
                     desc=None, model=None, participants=None,
                     smoothing=None, drop_missing=False,
-                    estimator=None, errorts=False,
+                    estimator=None, errorts=False, drift_model=None,
                     base_dir=None, name='fitlins_wf'):
     from nipype.pipeline import engine as pe
     from nipype.interfaces import utility as niu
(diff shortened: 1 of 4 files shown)


effigies (commented):
Does `cosine_*` work now? That would be how I would code it to ensure that I got all of the regressors, and I think the lack of support for that in PyBIDS was the reason I wasn't putting it in my models.

fitlins/workflows/base.py
```diff
@@ -85,6 +85,8 @@ def init_fitlins_wf(database_path, out_dir, analysis_level, space,
         iterfield=['session_info', 'bold_file'],
         name='design_matrix')
 
+    model.inputs.drift_model = drift_model
```
effigies, line 88:
```suggestion
    design_matrix.inputs.drift_model = drift_model
```

Though you could also just make it:

```Diff
     design_matrix = pe.MapNode(
-        DesignMatrix(drop_missing=drop_missing),
+        DesignMatrix(drop_missing=drop_missing, drift_model=drift_model),
         iterfield=['session_info', 'bold_file'],
         name='design_matrix')
```

adelavega:
@effigies This is fixed. If you sign off feel free to merge.

PR #280, opened 2021-01-28
FIX: AFNI TSNR calculation, ADD: BlurToFWHM,  ADD: testing outputs against reference

Shotgunosine:
Sorry about the size of this one, the number of changes kinda of got away from me. 
First big part of this is that I've fixed the TSNR calculation in the AFNI estimator. 
Second part is I added a new option to the AFNI estimator to use the 3dBlurToFWHM.
Third big part is that I've created a set of reference outputs for three different run settings (https://gin.g-node.org/shotgunosine/fitlins_tests). I also added estimator and smoothing information to the dataset description, but I can pull that out once @adelavega's PR is merged to master.


adelavega:
@Shotgunosine Go ahead and remove the `Smoothing` and `Estimator` writeout, I've merged the other PR.

Is this ready to test? I could try to run a model w/ neuroscout-cli. I think it would only test the first part (run normal model w/o smoothing).

adelavega:
Actually I could probaby also test `isoblurto`. Would this be a better default if using AFNI?

Shotgunosine:
I think you could probably test this tonight if you wanted to, but I should have the tests I wrote all passing by tomorrow morning. Isoblurto is pretty slow because it iteratively applies smoothing until the target smoothness is reached in all dimensions, so I don't think it's a good default. A great tool to have when some of your data has different smoothness, but not needed if everything's got about the same smoothness.

adelavega:
Okay cool if I get a chance I'll run it otherwise the tests should do it

adelavega:
I'm running a workflow with this PR right now

adelavega:
The workflow with this PR ran for me successfully. Otherwise, PR looks pretty good to me. Not sure I can review the TSNR stuff appropriately though.

Shotgunosine:
Alright, at this point I'm just not sure why the coverage is not properly capturing the tests on the new smoothing code. If anyone has any advice, I'd appreciate it, otherwise, this is ready for review.

adelavega:
You can run pytest with an "append" mode. That might help. see: https://github.com/PsychoinformaticsLab/pliers/blob/master/.github/workflows/python-package.yml

Key is `--cov-append`

Shotgunosine:
I don't think that should be it, each of the three tests (afni_smooth, afni_blurto, and nistats_smooth) is being run in a separate job on circleci, so they are each uploaded to codecov separately. As far as I can tell from the individual xml reports, they indicate that the lines have been run.

adelavega:
Hmm, I'm not sure then sorry.

adelavega (approved):
Otherwise PR LGTM. 
Overall LGTM.
 
From a high level only quibble I have is that the three test conditions differ in estimator and smoothing, but the latter seems less important overall to fitlins. Ideally we would test more permutations of fitlins parameters. That said, that seems like more of a reach goal, and this is already a nice improvement on what we have.

.circleci/config.yml
```diff
@@ -76,19 +76,15 @@ jobs:
...
+          command: |
+            datalad install -r -s https://gin.g-node.org/shotgunosine/fitlins_tests
+            datalad update --merge fitlins_tests
+            datalad get fitlins_tests/ds003/ds003_fmriprep/sub-0{1,2,3}/func/*_space-MNI152NLin2009cAsym_desc-*.nii.gz \
+                                    fitlins_tests/ds003/ds003_fmriprep/sub-0{1,2,3}/func/*_desc-confounds_*.tsv \
+                                    fitlins_tests/ds003/ds003_fmriprep/dataset_description.json \
+                                    fitlins_tests/ds003/ds003_fmriprep/sub-*/*/*.json
+            datalad get -r fitlins_tests/ds003/nistats_smooth/ fitlins_tests/ds003/afni_smooth/ fitlins_tests/ds003/afni_blurto/ 
       - save_cache:
          key: ds003-v0-{{ .Branch }}-{{ .BuildNum }}
```
effigies, line 89: We'll want to update all `ds003-v0` lines to `ds003-v1` to keep cache sizes down and make sure we're not depending on data from the old fetch that is no longer being kept fresh.

.circleci/config.yml
```diff
@@ -193,40 +191,38 @@ jobs:
...
-                --rcfile /src/fitlins/docker/multiproc.coveragerc \
-                $CONDA_PREFIX/bin/fitlins /data/sourcedata /out dataset \
-                -d /data -m /models/model-001_smdl.json -w /scratch \
-                --participant-label 01 02 03 --space MNI152NLin2009cAsym \
-                --estimator << parameters.estimator >> \
-                --n-cpus 2 --mem-gb 4 -vv
+                --rcfile /src/fitlins/fitlins/tests/multiproc.coveragerc \
+                $CONDA_PREFIX/bin/pytest --cov=fitlins \
+                --cov-config /src/fitlins/fitlins/tests/multiproc.coveragerc \
+                --cov-report xml:/scratch/.coverage_pytest_<< parameters.test_name >>.xml \
```
effigies, line 209: Seems a bit odd to `coverage run pytest --cov`. Should this just be `pytest`?

fitlins/interfaces/afni.py
```diff
@@ -80,16 +84,62 @@ def _run_interface(self, runtime):
...
+        # Need to do smoothing before scaling
         smoothing_fwhm = self.inputs.smoothing_fwhm
+        smoothing_type = self.inputs.smoothing_type
         if not isdefined(smoothing_fwhm):
             smoothing_fwhm = None
+            # Get input scans for 3dREMLfit
+            # If there's not smoothing, then the input bold file will do
+            img_path = self.inputs.bold_file
+            img = nb.load(img_path)
+        elif isdefined(smoothing_type) and smoothing_type == 'isoblurto':
```
effigies, line 100: This seems odd to do internal to the node. Is there a reason that we can't create a new node that does the smoothing in the workflow?

.circleci/config.yml
Shotgunosine: Part of trying to figure out why the coverage from the new tests isn't being correctly captured. It didn't help so it can likely revert to just pytest. I'm hopeful that the most recent commit will have resolved the coverage issues.

fitlins/interfaces/afni.py
Shotgunosine: Do you mean having a smoothing node prior to either estimator, or do you mean having the afni interface be encapsulated in a workflow with distinct smoothing and estimator nodes?

effigies: Just a node prior to the estimator. It should be skippable if the estimator will do the smoothing internally, e.g. with nistats and isometric smoothing.

Shotgunosine:
@effigies, I've finally got coverage working. Turns out the issue was pointing at tests in one directory (`/src/fitlins`) and calling fitlins from another location (`$CONDA_PREFIX/bin/fitlins`). This resulted in two copies of the source tree being in the coverage reports, one that had hits and one that didn't. For whatever reason, codecov interpreted that as meaning the lines had no hits. I changed the test location to point at $CONDA_PREFIX/lib/python3.6/site-packages/fitlins/tests and it's reporting coverage correctly now.

effigies:
Cool. It might be that we should have used `--pyarg fitlins` or something, but that can be fiddled with later.

I will try to review this today, but it's somewhat meeting heavy.

fitlins/interfaces/afni.py
```diff
@@ -339,6 +332,26 @@ def get_stim_labels(self):
         weights = _flatten([x["weights"] for x in self.inputs.contrast_info])
         return list(set(_flatten([x.keys() for x in weights])))
 
+    def save_tsnr(self, runtime, rbetas, rvars):
+        vol_labels = parse_afni_ext(rbetas)["BRICK_LABS"].split("~")
+        mat = pd.read_csv(self.inputs.design_matrix, delimiter="\t", index_col=0)
+        # find the name of the constant column
+        const_name = mat.columns[(mat != 1).sum(0) == 0].values[0]
```
effigies, line 339: Does this choke if there are non-steady-state volumes? Generally the constant should be zeroed out like everything else if you're censoring a volume.

fitlins/tests/test_outputs.py
```diff
@@ -0,0 +1,176 @@
...
+                if (rel_diff > max_rel).any():
+                    res.append(f"Relative difference (max of {rel_diff.max()})"
+                               f" greater than {max_rel} for {ref_nii} and {out_nii}.")
+            else:
+                if ((diff > max_abs).any()):
+                    res.append(f"Absolute difference (max of {diff.max()})"
+                               f" greater than {max_abs} for {ref_nii} and {out_nii}.")
+
+        else:
+            res.append(f"{ref_nii} nans don't match {out_nii} nans.")
```
effigies, line 163: Is this just

```Python
np.allclose(ref_data, out_data, rtol=max_rel, atol=max_abs, equal_nan=True)
```

effigies (commented):
On the whole looks good. Left a couple small comments.

fitlins/tests/test_outputs.py
Shotgunosine: I think it's different from numpy allclose. I was trying to mimic the functionality of [nibabel diff](https://github.com/nipy/nibabel/blob/95e1fbece92b92fc5a5f5c107bd0222438bdb0cc/nibabel/cmdline/diff.py#L183), but add in tolerance for nans. As far as I can understand nibabel diff handles relative and max specifications differently from numpy allclose. Allclose uses this:
`absolute(a - b) <= (atol + rtol * absolute(b))`

but nibabel diff uses the absolute tolerance as a mask and only tests the relative differences on elements where the difference is less than the absolute tolerance.

I'm not sure if those work out to be the same thing or not though.

fitlins/interfaces/afni.py
Shotgunosine: In addition to verifying that zeroing out everything else or not doesn't impact the estimates of the other coefficients, I also checked to see if we should expect a design matrix that doesn't have an all 1's constant column. We're getting the design matrix from nilearn's `make_first_level_design_matrix` function, which adds a an all 1s constant column as part of the drift model function, even if the drift model is None:
https://github.com/nilearn/nilearn/blob/d2ad41a9fa0aa7041c7e826518c6c3692a92a2e1/nilearn/glm/first_level/design_matrix.py#L122-L173

Based on that, I think this logic is fine.

PR #285, opened 2021-04-20
DBG: Pin nilearn 0.7.0 to check whether 0.7.1 changed statistics

effigies:


effigies:
@Shotgunosine Looks like https://github.com/nilearn/nilearn/pull/2567 changed the way z-scores are calculated. I'm inclined to update our baselines, since it looks like this was a well-considered change.

Waiting on tests to verify that this is in fact the cause.

Shotgunosine:
That seems likely to me given where the difference is. Looks like it's a spatially constrained set of ~ 400 voxels in a region that has z-scores near the threshold described in the issue that PR is based on. Since I already made it, here's a picture:
Screen Shot 2021-04-20 at 10 18 33 AM

I'll start running jobs to update the test datasets.

Shotgunosine:
Looks like blurto and smooth pass under 0.7.0, i'll update the test data on gin.

adelavega:
We found the nilearn part of the brain!

effigies:
> Looks like nistats blurto and smooth pass under 0.7.0, i'll update the test data on gin.

Sounds good. Will we need to update pins, or will tests just grab latest?

effigies:
> We found the nilearn part of the brain!

I would really enjoy an atlas of regions likely to be affected by upgrading a particular piece of software.

Shotgunosine:
@effigies I think it'll just pull the new data once it's uploaded.

PR #286, opened 2021-04-20
PIN: nilearn 0.7.1+

effigies:
Follow-up to #285 to require the new z-score behavior.


effigies:
> Relative difference (max of 2.0) greater than 0.0001 for /data/ds003/nistats_smooth/out/fitlins/sub-01/sub-01_task-r.../fitlins/sub-01/sub-01_task-rhymejudgment_space-MNI152NLin2009cAsym_contrast-taskVsBaseline_stat-FEMA_statmap.nii.gz

o_0

adelavega:
that's just a pure flake variation? woah

Shotgunosine:
No, that’s got to be a module version inconsistency or some other weirdness
local to the system where I ran it. I was trying to be quick and I just
used a compute node on our HPC where I already had a conda environment with
fitlins loaded. I made sure fitlins and nilearn were at the correct
version, but didn’t check the rest of the modules, and maybe more
importantly, didn’t check the work directory to make sure it was clean. I
was fairly confident that I was using fitlins and passing tests on that
environment on a previous fitlins version, but I’ll be a lot more confident
of the environment if I run it from a container. I’be built a docker
container off master and I’m going to rerun the test data generation with
that. If those results still manage to differ from CI by a factor of two,
then yeah, that’ll be a what the fuck.

effigies:
Thanks for updating the test data, @Shotgunosine!

PR #301, opened 2021-07-19
Exclude res from report entities

adelavega:
The entity key `res` causes a conflict in the report generation.

I'm simply excluding it from being passed on (currently running a model to confirm that this fix works)


adelavega:
Confirmed working for me

PR #322, opened 2021-11-01
FIX: Unreliable constant column name finding

### LINKED ISSUE #320: Logic to find constant column name fails due to floating point error
https://github.com/poldracklab/fitlins/blob/b55d28599efc28c9b0e08a797e3ebfb328e13367/fitlins/interfaces/afni.py#L339

This line is supposed to find the column in which all values are == 1.

Sometimes, presumably due to floating point rounding errors, this logic fails

For example, in one Neuroscout analysis id `bvymc` the constant column has values such as `1.0000000000000095` and thus fails this check. 

Seems to me that finding the constant column by value is a bit unreliable and instead we should do it by name, or only check values if a column named `constant` does not exist.

adelavega:
Fixes #320 by assuming constant name is: `constant`

### DIFF
--- a/fitlins/interfaces/afni.py
+++ b/fitlins/interfaces/afni.py
@@ -336,7 +336,10 @@ class FirstLevelModel(FirstLevelModel):
         vol_labels = parse_afni_ext(rbetas)["BRICK_LABS"].split("~")
         mat = pd.read_csv(self.inputs.design_matrix, delimiter="\t", index_col=0)
         # find the name of the constant column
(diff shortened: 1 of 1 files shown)


adelavega:
@effigies should be a pretty uncontroversial fix

adelavega:
@effigies should we also merge this or a similar fix to master?

effigies:
Yes this would be good. Looking at this now, I might use a better check:

```Diff
- const_name = mat.columns[(mat != 1).sum(0) == 0].values[0]
+ const_name = mat.columns[np.isclose(mat, 1).all(0)].values[0]

PR #329, opened 2022-01-10
ENH: Migrate CircleCI to github actions 

shashankbansal6:


.github/workflows/workflows.yml
```diff
@@ -0,0 +1,432 @@
...
+          #     allow-softlinks: true
+          #     show-channel-urls: true
+          #     use-only-tar-bz2: true
+
+          # - name: Check conda installation
+          #   shell: bash -l {0}
+          #   run: |
+          #     conda info
+          #     conda list
+
```
effigies, line 118:
```suggestion
```

.github/workflows/workflows.yml
```diff
@@ -0,0 +1,432 @@
...
+          cd $GITHUB_WORKSPACE
+          make html
+          ls
+          pwd
+
+      - name: Store artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: docs
+          path: docs/_build/html
```
effigies, line 197: Should we just use ReadTheDocs PR builds?

.github/workflows/workflows.yml
```diff
@@ -0,0 +1,432 @@
...
+        uses: codecov/codecov-action@v2
+        with:
+          files: /tmp/ds003/work/.coverage_pytest_${{ matrix.test_name }}.xml
+          flags: ${{ matrix.test_name }}
+          env_vars: GITHUB_JOB
+          verbose: true
+
+  deploy_version:
+    needs: [build_docs, test_ds003]
+    if: startsWith(github.ref, 'refs/tags/v')
```
effigies, line 327: I think this would require us to tag with `vX.Y.Z`, which we don't currently do.

```suggestion
    if: startsWith(github.ref, 'refs/tags/')
```

.zenodo.json
```diff
@@ -50,6 +50,11 @@
       "name": "Lee, John A.",
       "orcid": "0000-0001-5884-4247"
     },
+    {
+      "affiliation": "Stanford University",
+      "name": "Shashank Bansal",
```
effigies, line 55:
```suggestion
      "name": "Bansal, Shashank",
```

setup.cfg
```diff
@@ -18,16 +18,16 @@ description-file = README.rst
...
     nibabel>=2.0
     nipype>=1.3.2
     seaborn>=0.10.0
     numpy>=1.11
-    nilearn~=0.7.1
+    nilearn~=0.8.1
     pandas>=0.19
     tables>=3.2.1
-    pybids>=0.14.0rc1
+    pybids==0.14
```
effigies, line 30:
```suggestion
    pybids~=0.14.0
```

effigies (commented):
Overall looks good. Some small comments. I'm also going to try to replace your `neurodocker.sh` with an updated `.neurodocker.json`.

.github/workflows/workflows.yml
shashankbansal6: What is that? Like a separate workflow that runs on readthedocs PRs?

effigies: It's just something to enable in RTD. Unless there's a specific reason to have this in GitHub actions, I would move to that, as it's more predictive of whether we will be able to build the docs after a PR is merged. I'll go ahead and enable it and we can see.

effigies:
```suggestion
```

.github/workflows/workflows.yml
```diff
@@ -0,0 +1,416 @@
...
+      - name: Combine coverage and submit
+        uses: codecov/codecov-action@v2
+        with:
+          files: /tmp/ds003/work/.coverage_pytest_${{ matrix.test_name }}.xml
+          flags: ${{ matrix.test_name }}
+          env_vars: GITHUB_JOB
+          verbose: true
+
+  deploy_version:
+    needs: [build_docs, test_ds003]
```
effigies, line 310:
```suggestion
    needs: [test_ds003]
```

effigies (commented):
Can try removing `build_docs` and test the RTD PR builds now.

shashankbansal6:
> Overall looks good. Some small comments. I'm also going to try to replace your `neurodocker.sh` with an updated `.neurodocker.json`.

Yeah I was actually going to ask you to do this. Some of the stuff those neurodocker files are sort of patch-work so they definitely need your inputs.

PR #334, opened 2022-02-02
CI: Unify Docker deployment steps

effigies:
While I'm here trying to understand the build process for a little streamlining.


effigies:
Version stuff fixed. Merging so we can test the deployment.

PR #355, opened 2022-03-31
RF: Fitlins output reorganization

shashankbansal6:
Sorry, this took so long.

This is the first pass. I still have to move the reports to each node instead of creating them at the top level.

P.S. I think the pre-commit isort module made some minor changes to the libraries we are calling.


effigies (commented):
Quick thoughts. Node should always be defined, and is redundant with `level` and `name`.

PR #357, opened 2022-04-06
Allow models with no name

adelavega:
This PR handles models with a `Name` field set to "". This technically should not be allowed, but I think it's nice to handle this gracefully.

Eventually, a validator should catch this (although technically empty string is still a way to define this REQUIRED key, so its arguably valid).

### DIFF
--- a/fitlins/viz/reports.py
+++ b/fitlins/viz/reports.py
@@ -144,7 +144,7 @@ def write_full_report(report_dict, run_context, deriv_dir):
 
     model_name = snake_to_camel(report_dict['model']['name'])
     if model_name == "":
-        model_name = "unknown"
+        model_name = "Untitled"
     target_file = op.join(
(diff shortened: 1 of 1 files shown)


fitlins/viz/reports.py
```diff
@@ -142,9 +142,11 @@ def write_full_report(report_dict, run_context, deriv_dir):
 
     tpl = env.get_template('data/full_report.tpl')
 
-    model = snake_to_camel(report_dict['model']['name'])
+    model_name = snake_to_camel(report_dict['model']['name'])
+    if model_name == "":
+        model_name = "untitled"
```
effigies, line 147: We can even tolerate a missing `'name'` altogether:

```suggestion
    model_name = snake_to_camel(report_dict['model'].get('name') or "untitled")
```

adelavega:
@effigies I believe these test failures are unrelated, correct?

effigies:
Yup, pushed what I think is a fix for the test to master. Want to rebase?

PR #375, opened 2022-06-16
Adding a section on estimators

Shotgunosine:


docs/estimators.rst
```diff
@@ -0,0 +1,19 @@
+.. include:: links.rst
+
+Estimators
+==========
+Fitlins supports two estimators for first level models,
+Nilearn's :py:class:`nilearn.glm.first_level.FirstLevelModel`
+and AFNI's :ref:`3dREMLfit<afni:ahelp_3dremlfit>`.
+In both cases, the design matrix is generated by Nilearn.
```
adelavega: Well, I would be more specific and say it's generated by pybids + nilearn. pybids really does most of the work in design matrix generation actually.

docs/estimators.rst
```diff
@@ -0,0 +1,19 @@
+.. include:: links.rst
+
+Estimators
+==========
+Fitlins supports two estimators for first level models,
```
adelavega: We should also explicitly mention that Nilearn is the only option for higher-level models

adelavega (changes requested):
Good addition! Some comments

docs/conf.py
```diff
@@ -57,6 +57,11 @@
 
 source_suffix = ['.rst', '.md']
 
+intersphinx_mapping = {
+    'nilearn': ('https://nilearn.github.io/', None),
+    'afni': ('https://afni.nimh.nih.gov/pub/dist/doc/htmldoc/', None)
```
effigies, line 62: Huh. That's really cool.

docs/estimators.rst
effigies: Done in my changes.

effigies:
@Shotgunosine @adelavega What do you think? I'm not worried about the failing test. Seems one-off and other branches are passing.

Rendered version at https://fitlins--375.org.readthedocs.build/en/375/estimators.html

Shotgunosine:
@effigies, looks good to me!

effigies:
Thanks! And thanks for the first pass on this doc.

PR #378, opened 2022-06-17
DOC: Link out to BIDS Stats Models docs instead of maintaining our own

### LINKED ISSUE #377: Fitlins should point to https://bids-standard.github.io/stats-models/
FItlins bids stats-models page is out of date and should just point to https://bids-standard.github.io/stats-models/

Shotgunosine:
We probably don't want to duplicate the stats models documentation in the fintlins docs. Closes #377


Shotgunosine:
@effigies  and @adelavega do y'all have thoughts on this, #375 and #374?

effigies:
Do we actually cover everything in here, over there? If so, I think dropping `model.rst` altogether in favor of consistently linking to https://bids-standard.github.io/stats-models makes sense.

adelavega:
I think we do except the outputs section, perhaps we can leave that in? Otherwise LGTM

Shotgunosine:
@adelavega just getting back to this now. Which outputs section did you mean? Just this sentence? "The output of this level will again be a statistical map for the contrast a_vs_b, but summarized across the whole group."

@effigies, sounds good.

Shotgunosine (commented):
I don't know why, but switching from changes.rst to changes.md means that all of the version numbers get included in the toctree. I haven't been able to find a way to stop this
Screen Shot 2022-07-28 at 12 35 16 PM
.

adelavega:
I think that's all I meant but it doesn't seem very important now. I'm no help w/ the rst/md thing. 

Otherwise still LGTM

effigies:
@Shotgunosine Pushed a fix. Thanks for catching that.

effigies:
Failures unrelated. Diagnosing separately.

PR #381, opened 2022-07-22
ENH: Add stat entity to pybids config

effigies:
At some points we use level and name entities. I don't have the interest in figuring out why and pulling them out at this point, but we do also create files with the stat entity, so that should be added.


effigies:
Glad I didn't push this...

PR #383, opened 2022-07-28
DX: Pin pybids 0.15.1

effigies:


effigies:
Well, the good news is that the test failures are so easily fixed. I guess I'll bisect 0.15.1..0.15.2.

PR #386, opened 2022-09-07
CI: Add a build to test with unreleased pybids

effigies:


effigies:
@adelavega We're fixed on `pybids@master`.

PR #387, opened 2022-09-12
FIX: Handle correct contrasts and intercepts being passed from PyBIDS

effigies:


effigies:
@adelavega @jmumford @jdkent Looks like 0.15.3 runs, but now we're getting different values in our results. Suggests we're now using different design matrices, as nilearn and AFNI aren't changing, and neither are the input masks.

Do any of you remember changes that we would have expected to improve design matrix construction?

jmumford:
Could it be a change in the smoothing or highpass filtering options?

effigies:
Did we change those in PyBIDS? The only difference between this and the failing tests on the dev branch is that this is using pybids 0.15.1 and not 0.15.3.

adelavega:
Hmm. I guess I'd like to see a diff between the two DMs.

The only thing I can think of is that we were not modeling an intercept is some situations, and possibly we are adding an intercept where we shouldn't be? 

Unless there is a major bug (i.e. dropped columns), not sure where else the DMs would differ

effigies:
@adelavega Yes, the difference is that now there's an explicit `intercept` column. This didn't get overridden by nistats, which produced a `constant` column instead.

effigies:
FWIW I did write up my method for reproducing, since I wanted to get it set up on another system. https://github.com/poldracklab/fitlins/wiki/Testing-FitLins-regressions-on-example-data

adelavega:
Awesome, so that should be a fitlins issue then, correct?

effigies:
Well, it fixes the regression test, so that's good. Might have revealed other bugs in FitLins, though...

PR #411, opened 2025-05-14
fix: Find files without run entities

demidenm:
Modifying the entities list for cases where run (single run tasks) is not available in file paths. Tested on single run and multi-run case (ds000009). Sets runs entity to None when information is not available.

### DIFF
--- a/fitlins/interfaces/bids.py
+++ b/fitlins/interfaces/bids.py
@@ -361,6 +361,14 @@ class BIDSSelect(SimpleInterface):
         mask_files = []
         entities = []
         for ents in self.inputs.entities:
+            if 'run' in ents:
+                # Case: run is an integer (01, 1, etc.)
+                if isinstance(ents['run'], (int, float)):
+                    ents['run'] = [ents['run'], None]
+                # Case: run is a list with one element [1]
+                elif isinstance(ents['run'], list) and len(ents['run']) == 1:
+                    ents['run'].append(None)
+                    
             selectors = {'desc': 'preproc', **ents, **self.inputs.selectors}
             bold_file = layout.get(**selectors)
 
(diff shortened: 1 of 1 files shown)


effigies:
I don't 100% understand the problem this is trying to solve. Sorry if we discussed this separately. Could you show an example where the current behavior is bad?

demidenm:
Specifically:

> stats models implicitly adds a run-1 so that things can be split over run, and we take it too seriously when querying the layout in https://github.com/poldracklab/fitlins/blob/35f9a0d440731f3ad5630c56bda2780878885794/fitlins/interfaces/bids.py#L351-L403. So either we need to check for this case (maybe if entities['run'] == [1] set entities['run'] == [1, None]) or detect it when defining the spec in https://github.com/poldracklab/fitlins/blob/dev/fitlins/interfaces/bids.py#L268 and prune the run entity out.

fitlins/interfaces/bids.py
```diff
@@ -361,6 +361,14 @@ def _run_interface(self, runtime):
         mask_files = []
         entities = []
         for ents in self.inputs.entities:
+            if 'run' in ents:
+                # Case: run is an integer (01, 1, etc.)
+                if isinstance(ents['run'], (int, float)):
+                    ents['run'] = [ents['run'], None]
+                # Case: run is a list with one element [1]
+                elif isinstance(ents['run'], list) and len(ents['run']) == 1:
+                    ents['run'].append(None)
```
effigies: I think we can simplify, and we should leave a comment explaining why this is being done.

```suggestion
            # Hack around implicit "run-1" being added to files without run entities
            # This problem may be worth addressing in spec generation instead, but doing it here for now
            # See https://github.com/poldracklab/fitlins/pull/411
            if ents.get('run') in (1, [1]):
                ents['run'] = [1, None]
```

Could you check whether this patch works for your use case?

demidenm: Thanks. Will test on two separate datasets that do and don't include the edge case.

demidenm: @effigies - this worked in both test cases ds000009 and ds003858

PR #447, opened 2026-08-11
chore: General maintenance and CI recovery

effigies:
Triggering CI, making sure things still run.


effigies:
GitHub seems to be up for the moment and this got us everything but numeric equality on nistats_smooth back, and I'm guessing this is a minor API change from nilearn we can investigate. Separate problem deserves separate PR.

